PR #21, opened 2016-10-26
Show only vertical scroll

ai:
Some OS (Linux, Windows) still have ugly big scrollbars. Let’s remove horizontal scroll.

Before:

szq55m3

After:

hues6ga


EtienneLem:
Absolutely, thanks! I’ll also make sure that the padding don’t look twice as large on the right.

PR #22, opened 2016-10-31
Isolate .emoji-mart-anchors from page line-height

ai:
On our web site picker is looked like:

gkzyfht

(Note about spaces under category icons)

It happens because some of webpage styles inherits to component. Here we have a problem because of inherited `line-height` (we set it to 1.4).

This fix override problem.

(I wait until all browsers will support `all: initial`, so we will be able to do inherit properties isolation better).

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -31,6 +31,7 @@
   justify-content: space-between;
   padding: 0 6px;
   color: #858585;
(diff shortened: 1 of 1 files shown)


EtienneLem:
Thanks! I did consider `all: initial` too, but as you said browsers aren’t quite there yet.

PR #26, opened 2016-11-30
Allow a custom backgroundImageFn to be passed in

MarcoPolo:
This allows the emoji sheet to be served locally instead of always hitting the cdn. Useful for electron apps.

Also changes the github commit for emoji-data (aka emoji-datasource) to the npm's package. This saves a ton of time spent trying to git clone that repo.


EtienneLem:
Love it, thanks! 🤘 
Will merge and release later tonight.

PR #27, opened 2016-12-06
Allow custom namespace for store

anri-asaturov:
In multi-user environment (for example, Electron app), it would be nice for users to have separate 'frequently used' emoji lists.

This commits adds `setNamespace({string} namespace)` function to store module, so developer can set it to appropriate value after new user has been signed in.


EtienneLem:
Thanks! 🤘

PR #28, opened 2016-12-07
Allow passing children to Emoji component

savardc:
When use inside of an editor, it's useful to be able to pass children down to the Emoji component in order to be able to use the custom images and not break editing


EtienneLem:
Makes sense, thanks!

PR #29, opened 2016-12-13
Fix issue with operator in skin tone logic

savardc:
`||` has a higher precedence than conditional operator (`? ... :`). Fixed by surrounding with parentheses.


EtienneLem:
Woops, totally missed that PR. Thanks! 🤘

PR #34, opened 2017-01-15
wrap localStorage access in try-catch

trun:
Many browsers may support localStorage, but disable access to it in certain situations. For example, Safari private browsing mode will raise an exception if you try to access localStorage from a third-party domain, resulting in unhandled errors like...

> QuotaExceededError: DOM Exception 22: An attempt was made to add something to storage that exceeded the quota.

The easiest way to handle this for optional features is to simply try-catch all access to localStorage.


EtienneLem:
Thanks! 🤘

PR #35, opened 2017-01-18
Add emoji prop change check to Emoji component

scotato:
Hi, I ran into an issue with the Emoji component where the component would not update if the "emoji" prop changed, I added a rule to shouldComponentUpdate which resolves the issue.


EtienneLem:
Thanks! 🤘

scotato:
No problem!

PR #39, opened 2017-01-26
Category: export from index.js

Frugality:
(Accidentally added other commits to the branch source of this PR. Re-issuing with a branch containing only these changes).

Hi -

Really digging EmojiMart. It dropped in easily and works great!

I needed to do a little customization of the categories. Rather than modify the library directly, I decided to replace the methods at runtime, which necessitated exposing Category from the library.

Cheers,
-Courtland


EtienneLem:
Works for me, thanks!

PR #41, opened 2017-01-27
Preserve height and width on svgs

tracker-common:
We discovered an IE11 bug that would expand the height of SVGs. The solution was to not remove the height and width attributes from the SVG element.


EtienneLem:
Thanks! 🤘

PR #43, opened 2017-02-02
Add emojisToShow filter

tracker-common:
We would like to add an custom filter to only display certain emojis based upon their unified value.  For Windows 7/IE 11, anything with skin tones, flags and zero width joiners look pretty bad (i.e. their fallback values), so we want to filter that out using emojisToShowFilter property that gets passed in.

1. Search shouldn't show emojis that are filtered out.
1. Categories shouldn't show up if everything from that category is filtered out.  (For our case, we want to filter out flags)

Signed-off-by: Thomas Osugi <tosugi@pivotal.io>

### DIFF
--- a/src/components/category.js
+++ b/src/components/category.js
@@ -88,7 +88,7 @@ export default class Category extends React.Component {
   updateDisplay(display) {
     var emojis = this.getEmojis()
 
-    if (!display && !emojis) {
+    if (!emojis) {
       return
     }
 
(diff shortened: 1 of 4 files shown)


ELD:
We introduced an accidental regression with the first commit on this branch. Clearing out the search bar wouldn't result in showing all emoji again. The new commit fixes that regression.

EtienneLem:
Gotcha, thanks! Will force myself to find some time to get around to merging these PRs 👍

EtienneLem:
Works like a charm 👍  🤘

Issue #48, opened 2017-02-28
Data file unnecessarily large


vaskevich:
The npm package takes up just over half a meg, but most of it is actually the data file, which is about 500k. I was able to get this down to half the size (250k) by pruning unnecessary quotes "" around the keys and changing all the long key names to shorter, one letter keys.

Thoughts on changing the data file to make the package smaller?

vaskevich:
FYI, republished a light version here: https://www.npmjs.com/package/emoji-mart-lite

vaskevich:
@EtienneLem Cool! Although as it turns out, most of the reduction in that fork comes from getting rid of the sheet positions, the `has_img_XYZ` flags, and the same info for any `skin_variations`, which is needed for sheets to work.

There were other optimizations too (shorter keys, no quotes), but those didn't matter as much - when the JS bundle is gzipped (which many CDNs do), `emoji-mart` actually compresses pretty well to about 80k (vs 45k for `emoji-mart-lite`). I can send out a PR, tho!

EtienneLem:
Hey! Just published a new version with some improvements on the data file size (your PR + 2 other commits). The data file size went from 509 KB to 387 KB (not minified nor gzipped).

PR #49, opened 2017-03-02
Data file size reduction: strip quotes around keys

vaskevich:
https://github.com/missive/emoji-mart/issues/48

This seemed like the only file reduction change that wouldn't reduce additional complexity and brittleness, but take a look at `:master` and see if there is anything else you think would be worth including here.


EtienneLem:
Totally forgot to merge, thanks! 🤘

PR #50, opened 2017-03-04
use img for emojis to make them selectable and copyable using the alt attribute

stevenxxiu:
This commit uses img for emojis to make them selectable and copyable using the alt attribute. It helps a lot when the Emoji class is re-used for chat applications, so that users can copy an emoji instead of look it up. 

Correct me if I'm wrong but this sort of thing isn't possible with span as far as I know, I could at best manage to make the text within the span be copied, but the selection never highlights it.

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -15,7 +15,7 @@
 }
 
 .emoji-mart .emoji-mart-emoji {
-  padding: 6px;
+  margin: 6px;
 }
 
 .emoji-mart-bar:first-child {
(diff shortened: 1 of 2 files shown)


stevenxxiu:
@EtienneLem any updates? thanks.

EtienneLem:
Sorry about the delay. I’m afraid that we can’t merge this as-is, for this breaks native emojis (`children`) when supported. I remember having found a way to make the span selectable, I’ll dig into this.

vinnymac:
@EtienneLem @stevenxxiu yea it is possible to use a span I have done it in a private repo where we forked `js-emoji`. Just by adding the `shortname` as text to the span. Then you have to use CSS to hide the text with something like `text-indent: -9999px;`. I would show you the source if I could, but that is the trick I used and it works out great.

PR #53, opened 2017-04-01
Added `include` and `exclude` props to filter emoji categories

madshargreave:
Added #5


EtienneLem:
Clean and simple, thanks! 🤘

madshargreave:
My pleasure!

EtienneLem:
Available in `v0.5.0`. Made it so the keys are lowercase (like the `i18n` object). Search results also take include/exclude into account.

PR #54, opened 2017-04-04
add prop to autofocus search input

vinnymac:
Title explains it. Let me know if you'd rather use `el.focus()` or want me to change anything.


EtienneLem:
`autoFocus` props is how I would’ve done too, thanks! 🤘

SuperAL:
@vinnymac need this `el.focus()`, need to decide when to focus the input, is there any way to achieve this now?

PR #64, opened 2017-04-19
Add default frequently used defaults (closes #63)

### LINKED ISSUE #63: Default "Frequently Used"
Slack has a set of default frequently used icons. On Emoji mart the first time you load the picker (or clear localstorage) the frequently used is empty. This is confusing because the "Frequently Used" icon is selected in the top-left corner in the menu bar.

screen shot 2017-04-19 at 11 35 22

JustinDrake:


EtienneLem:
Thanks for the idea! I updated the list a bit. Also made it so the defaults are only saved when adding a new emoji to the frequently list. It will also use the `perLine` to return exactly 1 line of emojis (up to 16 max).

PR #68, opened 2017-04-24
Smoother anchor bar transitions 👍

JustinDrake:


EtienneLem:
So after 7 months I’ve decided to leave it the way it is. I like that Emoji Mart is as “pure” as possible (hence the very undesigned white background everywhere) and without animation by default. It’s then pretty trivial to let developers do whatever they want on top of that with CSS. 👍

PR #73, opened 2017-05-05
Installation instructions & stylesheet corrections

nsarafa:
1. introduce npm install instructions. Can be inferred but if you're like me, the first you do when trying new tools is `command + f` npm install

2. `word-break` was set to an invalid parameter

3. Mismatched property value threw an error in my text editor, so i fixed it


EtienneLem:
Thanks! 🤘

PR #74, opened 2017-05-05
update components from React.PropTypes to PropTypes package

nsarafa:
while using emoji-mart i noticed a persistent warning message in my console nagging me to stop using `React.propTypes`, and use the `prop-types` packages instead. That shouldn't be an issue for React 15+ when using this package any longer


EtienneLem:
Thanks! 🤘

PR #78, opened 2017-05-15
Update to latest emoji-datasource

savardc:
Emoji 4 support through latest version of emoji-data

I switched the default sheetSize to 32px since the 64px is not part of the main package anymore


EtienneLem:
Been following and wanting to upgrade emoji-data for a while, but using the 32px file as the default will most likely make the emojis blurry.

That being said, we could still have the updated `emoji-datasource` as a dependency, but link to the right sub-package at https://github.com/missive/emoji-mart/blob/master/src/components/emoji.js#L129

What do you think?

savardc:
I reverted the default sheetSize since the local version of `emoji-datasource` is not use for images (unless you have a `backgroundImageFn`)

samkelleher:
This looks good to be merged?  Seperatly I have a fork in which I am working on splitting out the emoji-data from this lib so it can be 'injected' rather than bundled. This is useful for me since my application uses the emoji-data lib elsewhere and don't want to bundle it twice.

EtienneLem:
Looks good to me! Thanks 🤘

PR #80, opened 2017-05-17
Update react and remove react-addons-test-utils

savardc:
- Update to react 15.5 to remove warnings about PropTypes
- Remove dependency on deprecated react-addons-test-utils. Functionality is now in react-dom


EtienneLem:
Thanks! 🤘

PR #82, opened 2017-05-23
Update yarn.lock

sonicdoe:
The `yarn.lock` file was not updated in #74 bringing it out of sync with `package.json`.


EtienneLem:
Nicely spotted, thanks! 🤘

PR #83, opened 2017-05-25
Do not store recently used if 'recent' category is excluded.

samkelleher:
If the `recent` category is excluded, this PR ensures that the Picker respects this and does not write to local storage anyway.

Previously, recent usage would be written to storage even if the `recent` category was ignored.

### DIFF
--- a/src/components/picker.js
+++ b/src/components/picker.js
@@ -102,7 +102,7 @@ export default class Picker extends React.Component {
 
     let includeRecent = props.include && props.include.length ? props.include.indexOf('recent') > -1 : true
     let excludeRecent = props.exclude && props.exclude.length ? props.exclude.indexOf('recent') > -1 : false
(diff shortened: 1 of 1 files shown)


src/components/picker.js
```diff
@@ -87,6 +89,7 @@ export default class Picker extends React.Component {
     let includeRecent = props.include == undefined ? true : props.include.indexOf('recent') > -1
     let excludeRecent = props.exclude == undefined ? false : props.exclude.indexOf('recent') > -1
     if (includeRecent && !excludeRecent) {
+      this.hideRecent = true
```
EtienneLem: I think this should be the other way around, right?

samkelleher:
Thanks for checking my mistake @EtienneLem - all good now. 😊

EtienneLem:
Sweet, thanks! 🤘

PR #84, opened 2017-05-25
Allow preview emoji to be hidden.

samkelleher:
When no emoji is being hovered, the default preview is shown (currently `department_store`).

By setting the `emoji` argument to an empty string, this fallback can be disabled.

This means that something will only show when an emoji is hovered.

This means that when both `emoji` and `title` are set to empty strings, the footer area will be empty.


EtienneLem:
Thanks! 🤘

PR #87, opened 2017-05-28
Make custom emojis work on example page

sonicdoe:
In addition to making custom emojis work on the example page, only the first `<Emoji />` example is now updated with the clicked emoji. I thought this would be the best solution as custom emojis won’t work with the rest of the examples (besides, not all emojis support the skin tone modifier, too).

Aside from that, the code example next to the displayed emoji is now updated with the actual `emoji.id` (or `{…}` in the case of custom emojis).

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -199,7 +199,7 @@ class Example extends React.Component {
 <Operator>import</Operator> &#123;Emoji&#125; <Operator>from</Operator> <String>'emoji-mart'</String>
 <br />
 <br /><Operator>&lt;</Operator><Variable>Emoji</Variable>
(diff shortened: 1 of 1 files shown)


EtienneLem:
Ended up using most of your code (sorry was easier to do in a separate commit rather than merge your PR), but I did let clicking emojis update all examples. I liked to see the difference between the native and the current set.

Thanks!

sonicdoe:
No worries. Don’t forget to update the `gh-pages` branch with v1.0.1 😉

EtienneLem:
Right! Done 😃

PR #88, opened 2017-05-31
Text representation of emoji is different from emoticon

savardc:
Emoji-data provides the ascii representation of the emoji, available under `text`. The emoticons are in the `texts` field. 

From the README for the `texts` field: *Each ASCII emoji will only appear against a single emoji entry*. This guarantees that an emoticon will map to a single emoji. 

The current version uses the ascii representation when no emoticon is available for an emoji, which leads to invalid results when looking up an emoticon. For example, `:)` returns `:blush:` and `:slightly_smiling_face:` where there should only be a single match `:slightly_smiling_face:`

I kept the ascii representation in the data since it might be useful to render a text-only version of the emoji.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,11 +90,12 @@ Sheets are served from [unpkg](https://unpkg.com), a global CDN that serves file
   id: 'smiley',
   name: 'Smiling Face with Open Mouth',
   colons: ':smiley:',
+  text: ':)',
   emoticons: [
     '=)',
     '=-)'
   ],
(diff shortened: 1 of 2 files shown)


savardc:
Hi @EtienneLem, do you see any issues with merging this? I understand if you're busy, I just want to make sure I didn't miss anything
Thanks

EtienneLem:
Makes sense, thanks! 🤘

PR #93, opened 2017-06-23
Fix shift when scrollbar is rendered

sonicdoe:
Including the width of the browser scrollbar in the total picker width resolves this issue.

Before:

Before

After:

After


EtienneLem:
Aah, much better! Been willing to fix that for a long time now, thanks! 🤘

Issue #95, opened 2017-07-25
Default value for Search


chadoh:
It might be handy to have a `defaultSearchValue` prop, which will update the `Search` component right after mount. I'm happy to contribute this change; let me know if you think it fits what you're going for.

EtienneLem:
👋  I could see a lot of use cases for that, that would work for me! I _think_ it will be as simple as setting a `defaultValue` on the input and calling `handleChange` when `Search` has mounted.

Happy to merge a PR 👍

nolanlawson:
This could potentially also fix #232, as you could pass an empty string into the default value.

Heet1996:
Hello,

Can someone please help me how can I pass the default query?

PR #97, opened 2017-07-30
Search/defaultValue: first pass

chadoh:
Ref: #95

Getting this to render properly seems like it takes more effort than it may be worth. Additionally, this causes the following error several times on re-render (like, say, when checking the 'autofocus' checkbox on the demo):

    Warning: flattenChildren(...): Encountered two children with the
      same key, `Custom`. Child keys must be unique; when two children
      share a key, only the first child will be used.

I have so far been unable to figure out why my change causes this error. However, as re-rendering will not happen as much under normal circumstances, maybe it doesn't matter.

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -43,6 +43,7 @@ class Example extends React.Component {
       hidden: false,
       currentEmoji: { id: '+1' },
       autoFocus: false,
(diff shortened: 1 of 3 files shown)


EtienneLem:

Issue #96, opened 2017-07-25
Searching for Country Flags


chadoh:
I just added the Uruguayan flag to [my life calendar](https://entire.life/chadoh/week/2017-11-21), but searching for "uruguay" did not find its flag. I had to search for the country code, "uy", which I didn't know ahead of time.

I imagine that adding searchability for all the flags is as simple as adding the country names to an object in some file. If you point me toward that file, I can submit a PR.

EtienneLem:
You’re right, I’ve never gotten around to making these easier to search. When building our data (`build-data.js`), we merge data from [emoji-data](https://github.com/iamcal/emoji-data) and [emojilib](https://github.com/muan/emojilib) (for the keywords). Looking at our script and JSON file we use for the keywords, I can see how the flags don’t get much love.

`short_name`s don’t match! We look for an object named `flag-uy`, while emojilib uses `uruguay` as the key. emoji-data don’t even have `uruguay` anywhere in its data, so we can’t use that to find the right object.

Here’s what I think would work (in `build-data.js`):
When building a Flag emoji (i.e. line 36), instead of getting the keywords with `datum.short_name`, we try to find the right object with the native character: 🇺🇾 !
I think we can use `unifiedToNative(datum.unified)` and find the object matching `char: 🇺🇾` in `emojilib`

This has always bugged me, it’s so ugly 😆 
![screen shot 2017-07-25 at 10 11 07 am](https://user-images.githubusercontent.com/436043/28576528-207da6c6-7122-11e7-8d94-519ff48ec21e.png)
We could change flags `datum.name` for the country name while we’re there. `emojilib` might need some changes as well, I see it’s not consistent.

Happy to assist or give a stab at if the above doesn’t make sense at all to you 😅

chadoh:
Phew! I stumbled into something that's much harder than anticipated! 

🏃 💨

chadoh:
Just kidding, I'm still curious to look into it. I might have a chance to dig in and fully try to comprehend the problem later this week. Using Emoji Mart has saved me a ton of effort and made the Entire.Life experience much nicer, so I'm happy to give back by making it even nicer.

chadoh:
Research:

* emojilib uses names like `uruguay` for 🇺🇾  and `gb` for 🇬🇧 , which all seem to align with the names that Github uses
* emoji-data uses names like `flag-uy` and `flag-gb`, which all seem to align with the names that Slack uses

[...]

PR #99, opened 2017-08-06
Search country flags using country names

chadoh:
This addresses #96.

This works around current limitations/inconsistences in `emoji-data` and `emojilib` by matching between the two libraries on the emoji character itself, rather than relying on a perfect match between the two disparate "short name" fields.

See individual commit messages, especially for a927821, for extra info.

Demo of new functionality:

country-search


EtienneLem:
This looks great @chadoh, I’ll do a thorough review ASAP!

EtienneLem:
That’s perfect, thanks again @chadoh!

Issue #100, opened 2017-08-14
Parse colons to unicode


Luchanso:
Hey guys, I need parse string with colons emoji like this: `"Hello :sunny:"` to `"Hello ☀️"`. I know how to parse colons:
```js
const colonsToUnicode = (text) => {
  const colonsRegex = new RegExp('(^|\\s)(:[a-zA-Z0-9-_+]+:(:skin-tone-[2-6]:)?)', 'g');
  let newText = text;

  let match;
  while (match = colonsRegex.exec(text)) {
    let colons = match[2];
    let offset = match.index + match[1].length;
    let length = colons.length;

    console.log(colons, offset, length);
  }
}
```
But I don't know how to convert in emoji. Maybe you can export [utils.getData(...)](https://github.com/missive/emoji-mart/blob/c21e1402b64e9f10f0a0eebb67f93791d654f06d/src/utils/index.js#L52-L124) for me? I think it the best solution :smiley:

Luchanso:
And maybe you can tell me how to parse Unicode symbols into emoji?

savardc:
You can use `emojis` from `emojiIndex` to convert from shortname to emoji

EtienneLem:
Once you’ve isolated the `colons`, you can either use the `Emoji` component (`<Emoji emoji={colons} />`) or as @savardc suggested above.

satyapendem:
You can use `emojiIndex` In emojiIndex just replace the colons string and give that string to eomjiIndex. 
Example: `emojiIndex.emojis[string]` it will give complete object of the particular emoji.

PR #101, opened 2017-08-14
Add exporting utils

Luchanso:
I'm exported utils, because I need getData(...) function, more information about it in issue #100


Luchanso:
I can't build it

PR #102, opened 2017-08-16
Add exporting utils

Luchanso:
I'm exported utils, because I need getData(...) function, more information about it in issue #100


EtienneLem:
I think this PR went a little beyond the original use case 😅. I think there are already few ways to handle your use case

PR #103, opened 2017-08-16
Add package size profiling tools

ai:
I love EmojiMart. But current library size is 110 KB (gzipped and minified). IT is too big (it is half or my app bundle size).

I want to help with size optimization. But every optimization must be started from profiling.

I added two tools:

* [size-limit](https://github.com/ai/size-limit) to prevent size regression. It will be run after test to check library size. MobX, JSS and Material UI use it too.
* [webpack-bundle-analyzer](https://github.com/th0r/webpack-bundle-analyzer) to see real content of package.

This tools always showed few ways for optimization

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,2 +1,3 @@
+dist/report.html
 scripts/
 .*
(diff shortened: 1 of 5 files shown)


ai:
@EtienneLem here is current content of `dist/emoji-mart.js`

j8nhbdb
 
When you will accept this PR I will send next PR to reduce this size

EtienneLem:
👍

Issue #105, opened 2017-08-18
Webpack bundle for Node.js?


ai:
For next steps of reducing size I need the decision from you.

![j8nhbdb](https://user-images.githubusercontent.com/19343/29388588-65853a16-829a-11e7-8798-119fe0504930.png)

Right now EmojiMart bundle everything including dependencies to webpack bundle at `dist/emoji-mart.js` and `dist/emoji-mart.min.js`. As result we have 3 problems:

1. The user will have 2 webpack preambles (webpack inner functions and magic) — one from their webpack, second from EmojiMart.
2. Even if the user has own `core-js` they anyway will have 2 `core-js` copies — one from their dependencies, second from EmojiMart. The main problem that `core-js` is really big. About 10% of the current EmojiMart bundle.
3. PropTypes slow down JS execution in production and keep size in the EmojiMart bundle. If the user has own PropTypes, they will have 2 copies of PropTypes.

My plan of size optimization:

1. During build process compile files by Babel to `dist/`, but don’t bundle them into the single file. `package.json`’s `main` will target to `dist/index.js`. Anyway, most of React users have webpack or other bundlers.
2. Bundle this files to `dist/emoji-mart.min.js` for backward compatibility, but do not use this file in `package.json`’s `main`.
3. `core-js` and all other dependencies will be in `package.json` `dependencies` and required with ` require('core-js')`. So we will use same `core-js` as the user already has.
4. Remove PropTypes with [`babel-plugin-transform-react-remove-prop-types`](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types).
5. Try to find code uses `process` and rewrite it, so webpack will not insert `process` polyfill. 

Question for @EtienneLem:

1. Do we really need `dist/emoji-mart.min.js`? I understand this file for jQuery plugins, etc. But it is extremely rare to use React without any bundler.
2. Do we really need `core-js`? Maybe we can rewrite code, that uses some ES6 classes? Or just put the warning to docs, if you need to support IE 11, please add `core-js` (for example, we do not send huge `core-js` to everyone, we send it only to IE 11 users).

ai:
On step 2 will optimize `data/index.js`. In `caniuse-lite` we found many ways to keep JSON size smaller.

ai:
@EtienneLem if questions are complicated, just tell me your concern. I will find extra data for decision.

EtienneLem:
Sorry for the delay and thanks for doing that research! 🤘 

---
[...]

PR #108, opened 2017-09-03
Feature/replace webpack tasks on babel

Muffassa:
Ref: #105 
I saved webpack for `build:example` and `test` tasks.


shaypeleg1:
What's the bundle size now after build?

Muffassa:
@shaypeleg1 107.25 KB

ai:
@Muffassa you forget to remove `webpack` from dependencies and remove unnecessary build scripts.

Also can you post here `npx size-limit --why` result? To be sure that PropTypes was removed.

Muffassa:
@ai Removed unnecessary build scripts. I tried remove `webpack`, but it required for building example build and running test.

ai:
@Muffassa at least you can remove `webpack-bundle-analyzer` since `size-limit --why` does the same

Muffassa:
@ai done

EtienneLem:
Awesome, thanks @Muffassa for contributing 🤘

sonicdoe:
I’m not sure I fully understand the reasoning behind removing `propTypes` in the build. As far as I can tell, I’ll now no longer benefit from `propTypes` during development. Shouldn’t [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) only be used by the application developer when doing a production build?

ai:
@sonicdoe theoretically, yes. But in practice, many developers don’t know about this `babel-plugin-transform-react-remove-prop-types`. We can add it to docs, but users don’t read docs.

But if type checking is really important, we can move this babel plugin to docs.

PR #111, opened 2017-09-11
Compress data

ai:
@EtienneLem More than 10% compression!

1. Data was moved from `data/index.js` to `data/data.js`.
2. Many fields were removed from `data/data.js`.
3. Uncompress code was added to `data/index.js` to add all fields removed in step 2.

As result API is the same (`require('emoji-mart/data')` return the same object), but size is smaller!


EtienneLem:
Impressive! Well done 🤘

PR #119, opened 2017-09-23
Use will-change:transform for smoother mobile scrolling

nolanlawson:
This improves mobile scrolling, [especially on mobile Chrome/Android](https://stackoverflow.com/a/41026886/680742). It provides a hint to the browser to move the scrollable element to a layer on the GPU, which leads to visibly smoother scrolling in Chrome/Android.

Chrome also warns about this in the Dev Tools:

out

Here's a video showing the before and after (note some overall lag due to using Android `screenrecord` plus DevTools being attached; the "after" version is even smoother without those things): https://gfycat.com/DamagedDistinctHorsechestnutleafminer

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -79,6 +79,7 @@
   overflow-y: scroll;
   height: 270px;
   padding: 0 6px 6px 6px;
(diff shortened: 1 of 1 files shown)


EtienneLem:
Thanks! 🤘

PR #120, opened 2017-09-24
Build as ES module as well as CommonJS

nolanlawson:
I can see this project is in active development, but I thought this PR may be useful to you, even just as a proof of concept.

In the [Mastodon](https://github.com/tootsuite/mastodon) project we are currently integrating `emoji-mart`, but we'd like to avoid the situation where we need to load the entire module (921kB) for all users. We use `emojiIndex` extensively, but the picker itself is only needed on-demand. So in principle, if `emoji-mart` were distributed as an ES module (with `"module"` declared in `package.json`), then we'd be able to do:

```js
import { emojiIndex } from 'emoji-picker'
```

...and [Webpack tree-shaking](https://webpack.js.org/guides/tree-shaking/) would take care of the rest.

I'm not sure if the current master is a good representation of what you intend to ship in `dist` (e.g. with raw `.svg` files?), but if so then this PR would be a reasonable approach to build both CommonJS and ES modules. Essentially it builds a `dist/cjs` directory (referenced by `"main"` in `package.json`) and then also a `dist/es` directory (referenced by `"module"` in `packagejson`). Webpack 2+ should pick up on the `"module"` version and thus allow for tree-shaking.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -12,7 +12,6 @@
     "transform-es2015-for-of",
     "transform-es2015-function-name",
     "transform-es2015-literals",
-    "transform-es2015-modules-commonjs",
     "transform-es2015-object-super",
     "transform-es2015-parameters",
     "transform-es2015-shorthand-properties",
@@ -40,5 +39,12 @@
         }
       }
     ]
-  ]
+  ],
+  "env": {
+    "cjs": {
+      "plugins": [
+        "transform-es2015-modules-commonjs"
+      ]
+    }
+  }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules/
 dist/
+dist-es/
 bundle.js
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "2.0.1",
   "description": "Customizable Slack-like emoji picker for React",
   "main": "dist/index.js",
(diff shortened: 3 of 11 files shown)


EtienneLem:
👋 Thanks for the contribution!

I understand and the Webpack tree-shaking sure sounds very interesting. I’m no Webpack pro, but given EmojiMart `src` files are already using `import` & `export`, would it work just by setting `"module": "src/index.js"`?

nolanlawson:
Sort of. If you do that, then your consumers would have to understand that when they use `dist` then they don't need any webpack loaders or babel plugins, whereas for `src` they do. I believe the optimal is to pre-apply the SVG-to-JSX transformation, Babel transformations, etc. and then produce both a CJS and an ESM build.

nolanlawson:
OK, I cleaned up and rebased this PR so that it writes to a new `dist-es/` directory, to maintain backwards compat for people who may already be using `dist/` in emoji-mart 2.x.

Incidentally, having an ESM build also makes the uglified webpack build slightly smaller:

```
648K	out-cjs.min.js
644K	out-es.min.js
```

You can test using e.g. `webpack --entry path/to/emoji-mart/dist/index.js --output-filename out.js && uglify-js -mc < out.js > out.min.js`.

EtienneLem:
I understand, makes sense! Thanks 🤘

PR #123, opened 2017-09-27
Upgrade to React 16

### LINKED ISSUE #122: Compatibility with React 16
>Error: Element ref was specified as a string (search) but no owner was set. You may have multiple copies of React loaded. (details: https://fb.me/react-refs-must-have-owner).

The code uses string-based refs in multiple places, I believe. That functionality was marked for deprecation in React 15. React 16 was just released, and no longer supports this feature.

Gargron:
Fix #122

- Upgrade React to 16.0.0
- Replace deprecated ref syntax
- Upgrade Webpack to 3.6.0
- Upgrade minor dependencies that don't have breaking changes

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,11 +1,44 @@
 {
-  "presets": ["es2015", "react"],
+  "presets": ["react"],
   "plugins": [
+    "check-es2015-constants",
+    "transform-es2015-arrow-functions",
+    "transform-es2015-block-scoped-functions",
+    "transform-es2015-block-scoping",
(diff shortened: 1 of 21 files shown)


Gargron:
I don't know if `measure-scrollbar` should be in `dependencies`, isn't the final product compiled so no runtime dependencies are needed?

EtienneLem:
Thanks! 🤘 

> I don't know if `measure-scrollbar` should be in `dependencies`

Mmh, I guess you’re right! I’ll check that out.

PR #124, opened 2017-09-27
Add a storybook

Gargron:
- Stories for Picker and Emoji components
- Knobs for easily adjusting those
- Live reloading of components makes development easier


EtienneLem:
That’s amazing, very useful. Thanks! 🤘

PR #130, opened 2017-10-06
Fix tests

nolanlawson:
The tests are currently broken, but this PR fixes them.

- Builds data to `src/data/data.js`
- Checks the `size-limit` based on the slightly smaller `dist-es/index.js`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 dist/
 dist-es/
 bundle.js
+/src/data/data.js
--- a/package.json
+++ b/package.json
@@ -62,12 +62,12 @@
     "webpack": "^3.6.0"
   },
   "scripts": {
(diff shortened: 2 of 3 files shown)


nolanlawson:
FWIW you might want to add Travis CI to catch these kinds of things in the future. 🙂

EtienneLem:
Definitely, thanks for doing that. Meant to do that right after merging your previous PR and release an update, but got caught up in work 😅 

I also wanted to investigate the little size regression in `dist`, my guess was that it could be 81f39b138a65ee9e0aa6bdd6d6f8ce0e5f413265 but I’m not sure.

Will try to merge & release update ASAP.

EtienneLem:
Thanks again! Doing a release as we speak 🤘

PR #132, opened 2017-10-07
Allow passing down a `recent` prop instead of localStorage

### LINKED ISSUE #129: Customize store backend
Ability to opt-out of localStorage and use something else instead. I am using Redux with my own persistence logic.

Gargron:
Fix #129 - Since the only things the store is used for is the "recent emojis" functionality and skin modifier persistence, if you pass down recent emojis through props and showPreview=false, the store will never be hit.


nightpool:
Since this adds a new property, it needs README documentation, right?

nolanlawson:
LGTM, I appreciate that it just doesn't even write or read from localStorage at all if you pass in `recent`. It's conceivable users would want to handle this themselves or use some other store (IndexedDB, etc.).

PR #133, opened 2017-10-09
Feature filter invalid emojis

Gargron:
This affects both localStorage and `recent`-backed "frequently used" category. Those store emoji IDs. But what if an emoji is deleted? The whole picker crashes. This ensures that invalid IDs are ignored.


dijonkitchen:
Related to #134 ?

PR #135, opened 2017-10-17
Update emoji-mart.css

fredvollmer:
For flexbox to work properly in IE10, the flex items need to have `display: block`. Additionally, the `flex` shorthand should be set explicitly, as IE10 has different defaults.


fredvollmer:
Can this get a review at some point?

EtienneLem:
Sorry for the delay on that one, thanks for opening the PR.
I did use your CSS but there was a lot of what looked like auto-styling included as well, so I just took the needed parts 🤘

PR #147, opened 2017-12-06
Support for Unicode 10.0 / Emoji 5.0

Vierkantor:
By updating the dependency on emoji-datasource, we can use newly introduced Emoji 5.0.

This required a change in emoji categories' names, but apart from that no big work.


EtienneLem:
Thanks! 🤘

PR #153, opened 2017-12-20
Fix search with custom data and filter

savardc:
`data.emojis` doesn't include custom emojis and is passing `undefined` to `emojisToShowFilter`


EtienneLem:
Thanks! 🤘

PR #155, opened 2017-12-28
add required props in <Emoji /> example

veob:
Without `size` `<Emoji />` doesn't render correctly.


EtienneLem:
Correct! Thanks 🤘

Issue #157, opened 2018-01-09
Unsupported emojis should render the shortcode


ajbeaven:
I've found that :shrug: doesn't seem to render anything when using the messenger set:

```
<Emoji
    emoji={':shrug:'}
    sheetSize={32}
    size={32}
    set="messenger"
    backgroundImageFn={(() => require('../../../assets/img/emoji-messenger-32.png'))}
/>
```

Presumably this is because the messenger set doesn't support this emoji? In my opinion, it's better to render _something_ rather than nothing. I'd suggest using the raw shortcode passed in to the emoji.

savardc:
We're also having this issue. I'm ok with rendering the shortname or maybe add a callback?

EtienneLem:
Went the callback way, as picker still requires unsupported emojis to be ignored. Here’s how you would render `:shrug:` when unsupported:

```js
<Emoji
  set={'messenger'}
  emoji={'shrug'}
  size={24}
  fallback={(emoji) => {
    return `:${emoji.short_names[0]}:`
  }}
/>
```

I hope that’s good enough 😄

ajbeaven:
Woah, that was crazy fast. Thanks heaps!

PR #158, opened 2018-01-09
Emoji fallback shouldn't apply styling #157

ajbeaven:
My use of the fallback mechanism just added is the same as was mentioned in the linked issue.

```
<Emoji
  set={'messenger'}
  emoji={'shrug'}
  size={24}
  fallback={(emoji) => {
    return `:${emoji.short_names[0]}:`
  }}
/>
```

In a case where the fallback is utilised, I'd prefer that styling wasn't applied.

Of course, feel free to close this pull request if there are other reasons that the font-size needs to be maintained!

### DIFF
--- a/src/components/emoji.js
+++ b/src/components/emoji.js
@@ -198,7 +198,6 @@ Emoji.defaultProps = {
   onOver: () => {},
   onLeave: () => {},
   onClick: () => {},
-  fallback: emoji => {},
 }
 
 export default Emoji
--- a/src/components/emoji.js
+++ b/src/components/emoji.js
@@ -123,7 +123,7 @@ const Emoji = props => {
 
     if (!setHasEmoji) {
       if (props.fallback) {
-        children = props.fallback(data)
+        return props.fallback(data)
       } else {
         return null
       }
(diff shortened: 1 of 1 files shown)


ajbeaven:
Ah, dang, it's because the fallback is rendered inside the emoji `<span>`. How about this...?

EtienneLem:
I see, it does make sense. Anyone wanting to apply a font-size (or any other style) can do so by styling whatever `fallback` returns (i.e. it can be JSX).

I think with that version we could remove the `fallback: emoji => {}` `defaultProps`. It really is my fault, it shouldn’t have been defined to being with, but we might as well make it part of that PR.

ajbeaven:
Sure thing! I've now removed the fallback from `defaultProps` (do participants get notifications of commits being added to PRs?).

EtienneLem:
Sweet, thanks!

> (do participants get notifications of commits being added to PRs?).

I do via email. Here’s my GitHub [notifications](https://github.com/settings/notifications) settings and I believe these are the defaults.

screen shot 2018-01-10 at 4 51 43 pm

Issue #166, opened 2018-02-18
Shared state between two pickers


akxcv:
![emojimartbug](https://user-images.githubusercontent.com/17192674/36439850-23d57b3e-167f-11e8-95d4-d9a49fcf69c0.gif)

When using two pickers, selecting an emoji in one picker affects another's state. How hard would it be to fix that? I understand that some state (e.x. frequently used emojis and default skin tones) should be shared between all pickers, but I feel like that is the one case where the state should not be shared.

EtienneLem:
Interesting 🤔 😅

I think I have a pretty good idea about what’s going on, I’ll check if there’s an easy fix.

PR #167, opened 2018-02-22
Custom Emojis as a changing prop

vcervellera-turbulent:
👋 

**Encountered issue:**
We provide the picker with custom emojis based on different contexts, which means that in certain circumstances, the picker has custom emojis that will not be present at the next opening.
However, the `CUSTOM_CATEGORY` emojis array is built outside of the `Picker` class, and is kept in memory for all uses of the Picker, and therefore keeps any emojis its provided in memory, even when the Picker receives an empty `custom` prop in the future.

**Fix:**
Upon construction of the `Picker`, as we check if the `custom` prop has no length, I check if the `CUSTOM_CATEGORY` object contains emojis with a length, and reset it if so, to avoid contamination.

### DIFF
--- a/src/components/picker.js
+++ b/src/components/picker.js
@@ -61,6 +61,8 @@ export default class Picker extends React.PureComponent {
       })
 
       allCategories.push(CUSTOM_CATEGORY)
+    } else if (CUSTOM_CATEGORY.emojis.length > 0) {
+      CUSTOM_CATEGORY.emojis = [];
     }
 
     this.hideRecent = true


savardc:
Hi @EtienneLem, just making sure you haven't missed this fix. Understand if you're busy. Thanks!

EtienneLem:
As a FYI: I noticed that we had other issues (#166) caused by the same thing. So although I did merge the PR, I sorta reverted it in the next commit (fed9a89) as it wasn’t required anymore.

These constants are now defined on each instance, instead of being shared. 😄

vishalDhull0910:
hey @vcervellera-turbulent can you please help me out here ,facing similar issue

Issue #86, opened 2017-05-26
Split out emoji library from emoji picker. 👻


samkelleher:
Just want to get a comment form the emoji-mart contributors.

I'm going to work on a fork that will split out using Higher Order Components the _rendering_ of the emoji-picker and the datasource.

At present, they are all bundled together, the picker imports directly the generate `data.js` file. However, I believe this complicates the the release cycle as often the underlying dataset will change more than the picker does.

Secondly, in my use case, the application uses the emoji dataset elsewhere. For example, in an rich text editor autocomplete. And in string replace to replace emoji with images when viewed on a older platform, or Windows.  So it doesn't make sense to have two copies of the emoji library.

Since making the changes required would likely result in an API change I wanted to get a feel for comments first before I started, especially related to component naming.

Something like:

Works as normal, uses localstorage and internal data library.
```
<Picker />
```

`Picker` would just wrap the `SimplePicker` and provide it with the internal data and localstorage functions. However, my application could consume the lower level component directly, which I've called `SimplePicker` which would expect the storage/dataset functions to be injected into it.
```
const data = application.loadEmojiDataLib();
const saveSkinTone = skinTone => application.storeUserPreference({ skinTone });

<SimplePicker data={data} onSkinTonePreferenceChange={saveSkinTone} />
```

This means that by referencing only the SimplePicker, webpack would not bundle the included emoji dataset in my application code.

Thoughts / comments? 💬

savardc:
We also use the dataset separately from the picker and having an easier way to query it would be a big improvement

EtienneLem:
That would work for me 😄

slvnperron:
@samkelleher Have you done that? I also need to split data from the bundle

samkelleher:
@slvnperron I did actually and finished the other day, however, by the time I finished and moved things about I have something thats radically different to emoii-mart. Visually similar, but by time I had moved stuff around and generally optimised and tweaked it I essentially have something new.  Need to figure out where to take it from here.

mismith:
What's the status of this?

samkelleher:
@mismith I need to figure out what to do since when I split out the emoji dataset to essentially recreate an emoji picker using dumb components decoubled from their dataset, it obviously is a breaking change to emoji-mart so there is no clear upgrade path or replacement. I could continue to refactor and try preserve the original interface but this would take a lot of work and I've already completed the part I needed.

mismith:
Gotcha, thanks for the clarification.

[...]

PR #173, opened 2018-03-12
Add parent components and data property for components that reference data.js Issue#86

davidcilley:
This pull request is an attempt to create all of the original emoji-mart component behavior, while decoupling it from the large dataset contained in data.js.

This adds a parent-like component prefixed with "Nimble" for the following objects:
**Picker, Category, Emoji, Preview, Search**

The parent components now have a required data property so that a partial or changed version of the original data.js output can be used. The child components pass the original data.js output to maintain backwards compatibility.

This also updates several of the util methods to allow for data passage.  One possible breaking change, "emoji-index.js" now has an instanced version of the original globals, and uses a new instance of NimbleEmojiIndex with data.js as it's input.  Maybe this addresses PR #172 ?

I attempted to keep data.js from being referenced throughout all of the new Nimble components, but in my test app, WebPack was still outputting the original data, even though I was referencing the components directly. This may still need some tweaking to fully address  Issue #86 

Unfortunately, I couldn't find a way to keep backwards compatibility, without moving all of the original code to the parent Nimble objects. This will remove a lot of the original git history (unless someone has a better idea!)


vcervellera-turbulent:
:wave: 
Thanks for mentioning the PR I opened, I just tested yours to see if it was patching our issue, and unfortunately it doesn't, there seems to be a custom emoji contamination on that PR as well.
I will create a jsfiddle to outline our issue and update my PR with it shortly, in case you want to give it a go. 🐝

davidcilley:
Sorry all, I need to close this PR and go through proper channels on my end. Got a little ahead of myself!

EtienneLem:
@davidcilley No worries, thanks for giving it a stab! Will be happy to review when you’re ready 👍

PR #175, opened 2018-03-15
Select first emoji in list with `enter`

savardc:
Add the ability to select the first emoji after a search with `enter`

I added a new props, `onSelect`,  to be notified when an emoji is selected, either with a click or `enter`. I'm not sure how much the mouse event is used so I didn't want to just rename `onClick` to `onSelect`. 

The next step would be to add keyboard navigation in the list of emojis


EtienneLem:
Very nice, thanks! 🤘

PR #187, opened 2018-05-08
Correct NimbleEmojiIndex reexport

Kovensky:
`./utils/emoji-index/nimble-emoji-index` has a default export, not a named export.


EtienneLem:
Whoops! Thanks 🤘

PR #203, opened 2018-07-10
Added font stack to support native emojis in IE11/Edge

luisbelloch:
Hi! This adds support for native emojis in IE11, previously rendered as black & white icons. Adjusting the font stack will also enable full support on Edge too, some emojis are broken too.

IE11 on Windows 10 using native emoji:

ie11-before

Same platform, but after applying the change, it should render native emojis properly:

ie11-after

I didn't test in in previous windows versions, I suspect it won't work if font is not available.

Emoji-mart is awesome, saved us a lot of time - thanks for contributing it to the world!

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -138,6 +138,10 @@
   font-size: 0;
 }
 
+.emoji-mart-emoji-native {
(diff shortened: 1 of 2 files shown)


EtienneLem:
Thanks! 🤘

PR #206, opened 2018-07-13
Not Found Component for emoji search

Rena-Ryumae:
If a user's search returns no results,
the sleuth_or_spy emoji appears. We would
instead like for a custom image to appear.

We propose creating a NotFound component
that uses the the sleuth_or_spy emoji
on default when a user's search returns
no results but otherwise uses what the
user passes through


EtienneLem:
Changes to note: I made it so when using a custom component, it doesn’t have the `emoji-mart-no-results` applied. That’s up to the developer to handle spacing / layout correctly when using a custom component.

As a bonus while in that area: I added a `notFoundEmoji` props to still use the default not found component, but a different emoji (default still is `sleuth_or_spy`).

PR #208, opened 2018-07-31
Customizable Category Icons

Rena-Ryumae:
Currently, the default emoji-mart
category icons cannot be changed.

We propose creating a categories
object that uses the emoji-mart
category icons on default but
otherwise uses custom imgs/svgs
passed in by the user


EtienneLem:
Thanks Rena! These updates are great 🤘 

`v2.7.0` has just been published.

PR #209, opened 2018-08-02
Custom emoji for skin tone selector

Rena-Ryumae:
Currently, the skin tone selector
is a row of colored dots

We propose creating a skinIcon field
that uses the emoji-mart dots on
default but otherwise uses the emoji
passed in by the user


EtienneLem:
That’s awesome, thanks! 🤘

PR #213, opened 2018-08-08
fix typo in README.md

tkbremnes:
The sheet size for messenger says that the emoji download size is 449
_megabytes_ for a sheetSize of 32.
449 _kilobytes_ is the correct value.


EtienneLem:
Well spotted 🕵️, thanks!

Issue #229, opened 2018-09-06
Reducing bundle size with NimblePicker doesn't seem to work with Webpack


rmccue:
Hi! First off, love Emoji Mart, it's an amazing component. We tried building our own emoji picker before and it a total pain, so having a great component like this available is fantastic.

We have an app using Webpack (by way of create-react-app's react-scripts), and I was hoping to be able to slightly reduce the bundle size by switching to NimblePicker [as mentioned in the docs](https://github.com/missive/emoji-mart#datasets).

Unfortunately, this does not appear to be working, as source-map-explorer shows that both `twitter.json` *and* `all.json` are now being included:

<img width="1241" alt="screenshot 2018-09-06 11 35 49" src="https://user-images.githubusercontent.com/21655/45129800-3923f200-b1c9-11e8-84db-d366eed93881.png">

(This is using the example from the readme, but with `twitter.json` rather than `messenger.json`)

From what I can see, this seems to be a limitation of webpack where they're following the spec which says [all child modules must be evaluated](https://github.com/webpack/webpack/tree/master/examples/side-effects) as they may contain side-effects.

I was able to get the bundle size down by importing directly from `dist-es/components/picker/nimble-picker` *and* modifying the `components/` files to use direct imports rather than `import { Skins } from '.'` (e.g.). Super annoying to do so though.

This appears to be true with webpack 4 as well; switching to NimblePicker + direct data import increases the built size by 63KB (source-map-explorer doesn't show the JSON files for whatever reason, so can't get a visual comparison). However, adding `"sideEffects": false` to the `package.json` **does** make this work correctly. (With that set, switching to NimblePicker reduces the bundle size by 2KB of the final size, which I think is expected.)

Anyway, long journey investigating this, but I'm thinking two changes could be made on the Emoji Mart side:

1. Add `"sideEffects": false` to `package.json` to support tree-shaking in Webpack 4
2. Document that the bundle size will not be reduced and in fact, will increase with Webpack <4 if using NimblePicker.

Happy to send through PRs for these two points if you're in favour!

jamessharp:
I ran into this too. Adding

```
"sideEffects": [
    "*.css"
  ]
```

to the `package.json` (of this library) sorted it out. Without the css string the css will get stripped by tree shaking.

EtienneLem:
You may all thank @rtkhanas, it should be working now (v2.9.0). See #245

nolanlawson:
[...]

PR #245, opened 2018-11-13
Import compoments directly to avoid components/picker/index.js and have smaller bundle

rtkhanas:
Those simple changes would allow importing Nimbleicker directly and avoid having all.json bundled when using apple.json or other emoji collections.

### DIFF
--- a/src/components/category.js
+++ b/src/components/category.js
@@ -3,7 +3,8 @@ import PropTypes from 'prop-types'
 
 import frequently from '../utils/frequently'
 import { getData } from '../utils'
-import { NimbleEmoji, NotFound } from '.'
+import NimbleEmoji from './emoji/nimble-emoji'
+import NotFound from './not-found'
 
 export default class Category extends React.Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


rtkhanas:
Probably related issue: https://github.com/missive/emoji-mart/issues/229

EtienneLem:
Interesting! I was under the impression that it was tree shaking’s job to be smart about these imports.

EtienneLem:
Thanks 🤘

rtkhanas:
I think tree shaking is not possible with old `require` and only works with `import`. I might be wrong.

Anyway, this is a super simple way to reduce bundle size with no tree shaking and other advanced things. No need to import a lot and strip useless code if we can just import what is needed. 😃 

Thank you for merging, would be nice to see a new `emoji-mart` version in NPM! 😊

EtienneLem:
> I think tree shaking is not possible with old `require` and only works with `import`. I might be wrong.

That’s my understanding as well, which is why we have 2 dist folders and specify the one with `import` via package.json `module`.

```json
"main": "dist/index.js",
"module": "dist-es/index.js",
```

But as far as I’m concerned we’ve been lied to… 😅 

> Thank you for merging, would be nice to see a new `emoji-mart` version in NPM! 😊

ETA few minutes 😄

rtkhanas:
**Before:**

screen shot 2018-11-13 at 11 15 36 pm

**After:**
screen shot 2018-11-13 at 11 50 54 pm

Thank you 😃

EtienneLem:
Amazing! Thank **YOU** 🤘

Issue #193, opened 2018-05-23
Need to convert native emoji in a string to colons


mukulbirdeye:

satyapendem:
I too have the same problem How I can do that please help me?

nolanlawson:
Could you specify what you mean by converting the native emoji to colons? I'm not sure I understand the use case.

Issue #16, opened 2016-10-18
Get Emoji from its unicode


iamsaksham:
is there any way by which we can get Emoji from its unicode?

EtienneLem:
Not at the moment

goodmind:
Is there any progress?

Fauntleroy:
This is the temporary solution I added for my application:

```js
import _ from 'lodash';
import { emojiIndex } from 'emoji-mart';

const { emojis } = emojiIndex;

export function getEmojiDataByNativeString (nativeString) {
  return _.find(emojis, { 'native': nativeString });
}
```

Not sure if this works in all cases yet. `lodash` isn't essential for this, I just use it elsewhere in my application.

stnwk:
Example without lodash:

```jsx
import { emojiIndex } from 'emoji-mart';

export function getEmojiDataByNativeString (nativeString) {
  return Object.values(emojiIndex.emojis).find(item => item.native === nativeString)
}
```

johnsenpeder:
Won't work when the unicode has has skin tones in it

jmcrthrs:
@johnsenpeder Very nice thanks.

However I have found a few instances where this doesn't work. They appear to be gender related:

person_with_ball
man-bouncing-ball
woman-bouncing-ball

https://codesandbox.io/s/oq8rkky5yq

johnsenpeder:
@jmcrthrs yeah, seems to be gender related... :thinking: interesting..

ifiokjr:
A quick fix would be to add the following since no other gender dependent emoji which I checked had the same problem. That being said I haven't investigated this fully.

```
import { emojiIndex } from 'emoji-mart';
import { BaseEmoji, CustomEmoji } from 'emoji-mart/dist-es/utils/emoji-index/nimble-emoji-index';

const isBaseEmoji = (emoji: BaseEmoji | CustomEmoji | undefined): emoji is BaseEmoji => {
  return Boolean(emoji && (emoji as BaseEmoji).native);
};

export function getEmojiDataByNativeString(nativeString: string) {
  const skinTones = ['', '🏻', '🏼', '🏽', '🏾', '🏿'];

  let skin = null;
  let baseNativeString = nativeString;

  skinTones.forEach(skinTone => {
    baseNativeString = baseNativeString.replace(skinTone, '');
    if (nativeString.indexOf(skinTone) > 0) {
      skin = skinTones.indexOf(skinTone) + 1;
    }
  });
  
[...]

PR #250, opened 2018-11-29
Get emoji data from native

pederjohnsen:
Adds function to get emoji data from native unicode emoji

This resolves https://github.com/missive/emoji-mart/issues/16 and resolves https://github.com/missive/emoji-mart/issues/193

**[edit]**
Thought I had already fixed it but seems the issues with some emojis as mentioned in https://github.com/missive/emoji-mart/issues/16#issuecomment-390960784 are still present here too.. If anyone has any thoughts as to why and how to fix it let me know :upside_down_face:


ifiokjr:
This looks very helpful. Any chance it gets merged soon?

nolanlawson:
It seems based on discussion in #16 that this still needs some more work?

pederjohnsen:
Yes unfortunately it still needs more work, I've not got the time to revisit this at the moment, but once I have (unless someone else has had a look by then) I'll post an update in here :ok_hand:

pederjohnsen:
@nolanlawson I've updated the code and it seems to work correctly for all emojis now...

This also includes an update to `nimble-emoji-index.js` which returns search results for the skin tone you've picked in the picker. If I remember correctly this isn't required for the `getEmojiDataFromNative` to work. But was another thing I needed for an app I'm working on. I could pull this out and create a separate PR if wanted :+1:

src/utils/index.js
```diff
@@ -134,6 +135,49 @@ function getData(emoji, skin, set, data) {
...
+  }
+
+  const skinTones = ['', '🏻', '🏼', '🏽', '🏾', '🏿']
+  const skinCodes = ['', '1F3FB', '1F3FC', '1F3FD', '1F3FE', '1F3FF']
+
+  let skin
+  let skinCode
+  let baseNativeString = nativeString
+
+  skinTones.forEach((skinTone) => {
```
nolanlawson: Since the first `skinTone` is an empty string, is a good idea to search for it here?

src/utils/index.js
```diff
@@ -134,6 +135,49 @@ function getData(emoji, skin, set, data) {
...
+    if (nativeString.indexOf(skinTone) > 0) {
+      const skinToneIndex = skinTones.indexOf(skinTone)
```
nolanlawson: We can figure out this `skinToneIndex` by using `skinTones.forEach((skinTone, i) => {}) instead

src/utils/index.js
```diff
@@ -134,6 +135,49 @@ function getData(emoji, skin, set, data) {
...
+      skin = skinToneIndex + 1
+      skinCode = skinCodes[skinToneIndex]
+    }
+  })
+
+  const emojiData = Object.values(data.emojis).find((emoji) => {
+    emoji = JSON.parse(_JSON.stringify(emoji))
```
nolanlawson: Let's consistently use `_JSON` here

src/utils/index.js
```diff
@@ -134,6 +135,49 @@ function getData(emoji, skin, set, data) {
...
+
+  skinTones.forEach((skinTone) => {
+    if (nativeString.indexOf(skinTone) > 0) {
+      const skinToneIndex = skinTones.indexOf(skinTone)
+      skin = skinToneIndex + 1
+      skinCode = skinCodes[skinToneIndex]
+    }
+  })
+
+  const emojiData = Object.values(data.emojis).find((emoji) => {
```
nolanlawson: `Object.values()` is not supported in all browsers, this will probably require a new babel polyfill to work. Can we just use a for-loop instead?

src/utils/index.js
```diff
@@ -134,6 +135,49 @@ function getData(emoji, skin, set, data) {
...
+
+    if (emoji.variations && emoji.variations.length) {
+        emoji.unified = emoji.variations.shift()
+    }
+
+    if (skin && emoji.skin_variations && emoji.skin_variations[skinCode]) {
+      emoji.unified = emoji.skin_variations[skinCode].unified
+    }
+
+    return unifiedToNative(emoji.unified) === baseNativeString
```
nolanlawson: It seems odd to me to clone the whole object just to do a lookup, this seems expensive. Is there another way to do this?

.editorconfig
```diff
@@ -0,0 +1,2 @@
+[*]
+indent_size = 2
```
nolanlawson: This is a good change but would also probably make sense in a separate PR. :)

pederjohnsen:
Thanks for reviewing, I'll try and get on this over the weekend 👍

pederjohnsen:
@nolanlawson I've updated the code and added some tests, let me know if you find anything else :ok_hand:

nolanlawson:
This LGTM, I think the only thing missing is that it would be great to have a Jest test for this, since this is a pretty self-contained little util function. I don't think it should necessarily block this PR from being merged, though.

pederjohnsen:
https://github.com/missive/emoji-mart/pull/250/commits/ab2f6004ac5d99e3823630bfe6f98870205e99d2 ? :thinking:

nolanlawson:
Oops, my bad, did not see the tests. Everything LGTM!

pederjohnsen:
Thanks :) :tada:

Issue #247, opened 2018-11-18
spritesheets for custom emojis?


alexpmorris:
we created a custom spritesheet just for custom emojis, however it doesn't seem there's an easy way to pass in the associated css class for them, as opposed to using `imageUrl`.  Is there some easy way we may have overlooked to accomplish this?  If not, perhaps this would be a nice addition for an upcoming release.

This is the tool we used to create our spritesheet: https://github.com/draeton/stitches

The following is a sample of the associated CSS:

```
.custom_emojis {
    background-image: url(custom_spritesheet.png);
    background-repeat: no-repeat;
    display: block;
}

.custom_emojis-custom_angry {
    width: 32px;
    height: 32px;
    background-position: -2px -2px;
}
```

Thanks for any guidance you can offer with this, and for this terrific widget as well! :)

EtienneLem:
Custom emojis only work with specific image URL at the moment, unfortunately.

Is using CSS classes important for your use case, or setting the position via the `custom` param works for you? The only thing missing would be to support `position: [row, column]` to custom emojis and EmojiMart would take care of the rest (like it currently does for built-in emojis).

alexpmorris:
The only reason for the CSS is positioning, so if the whole process could be streamlined internally by EmojiMart via the `custom` param that way, that should work great too!  

{ name: 'Blush',
      short_names: ['my_blush'],
      keywords: ['myblush'],
      spriteUrl: 'custom_spritesheet_32px.png',
      position: [row, column]
},

That would definitely make for a terrific addition!

alexpmorris:
great work @Kennybll, definitely speeds things along... and hopefully this can be merged in short order! 👍 😁

EtienneLem:
v2.9.2 has just been released 😄

PR #251, opened 2018-11-30
Adds Spritesheets for Custom Emojis

Kennybll:
Adds support for spritesheets as asked in https://github.com/missive/emoji-mart/issues/247. I opted out of using 'position: [row, column]' because the variables sheet_(x|y), size, and sheet(Columns|Rows) are already implemented. The main thing added is spriteUrl to the custom emoji data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,7 @@ Following the `dangerouslySetInnerHTML` example above, make sure the wrapping `s
 ```
 
 ## Custom emojis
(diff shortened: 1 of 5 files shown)


alexpmorris:
great work @Kennybll, definitely speeds things along... and hopefully this can be merged in short order! 👍 😁

EtienneLem:
Great work indeed. Exactly what I had in mind, thanks! 🤘

PR #258, opened 2018-12-18
fix: allow prop-types to be removed in production

### LINKED ISSUE #257: prop-types are not actually removed in production
Currently this package is using [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types) but it's not actually removing `prop-types` because the usage of `prop-types` is too subtle for the Babel transform to figure out.

Also, our usage precludes the possibility of library users from removing it themselves using this same Babel plugin.

You can reproduce this issue by running:

```bash
yarn build
```

Then:

```bash
egrep -r '(prop-types|shared-props)' dist dist-es
```

Then you'll see that there are several files importing these modules, even in the production build.

nolanlawson:
fixes #257 

There were a few things required in order to actually remove `prop-types` from the build:

1. Split `shared-props.js` into `shared-props.js` and `shared-default-props.js`, because one needs to be removed and the other one doesn't.
2. Use the [`additionalLibraries`](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types#additionallibraries) option in `babel-plugin-transform-react-remove-prop-types` to catch `shared-props.js` in addition to `prop-types`
3. Update the babel transform.
4. Use `removeImport: true`.
5. Add comment annotations to help give the babel transform some hints about what to remove.

I can confirm that this fixes the issue by running:

```bash
egrep -r '(prop-types|shared-props)' dist dist-es
```

And I observe that `prop-types` is only imported by `shared-props.js`, and `shared-props` is not referenced by any other module.

Now, this PR raises the question of whether we _actually_ want to unilaterally remove the `prop-types` from the npm package. If we do, then we lose the debugging benefits.

However, I tried using the `"mode": "wrap"` option of the babel transform (to wrap everything in a `process.env.NODE_ENV === "production"` check, and it ran into a runtime error for some reason.

So perhaps a better solution is to build multiple versions:

```
dist/
dist-es/
dist-prod/
dist-prod-es/
```

Alternatively, we could make these code changes but _not_ actually run the transform ourselves, but then provide instructions in the readme for configuring the transform. Then consumers of this library would be able to decide for themselves if they want to remove the `prop-types`.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -23,7 +23,6 @@
 
     "transform-object-rest-spread",
     "transform-runtime",
-    "transform-react-remove-prop-types",
     [
       "transform-define", "scripts/define.js"
     ],
--- a/README.md
+++ b/README.md
@@ -365,6 +365,32 @@ Apple / Google / Twitter / EmojiOne / Messenger / Facebook
 ## Not opinionated
 **Emoji Mart** doesn’t automatically insert anything into a text input, nor does it show or hide itself. It simply returns an `emoji` object. It’s up to the developer to mount/unmount (it’s fast!) and position the picker. You can use the returned object as props for the `EmojiMart.Emoji` component. You could also use `emoji.colons` to insert text into a textarea or `emoji.native` to use the emoji.
 
+## Removing prop-types in production
+
+To remove [prop-types](https://github.com/facebook/prop-types) in production, use [babel-plugin-transform-react-remove-prop-types](https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types):
+
+```bash
+npm install --save-dev babel-plugin-transform-react-remove-prop-types
+```
+
+Then add to your `.babelrc`:
+
+```json
+"plugins": [
+  [
+    "transform-react-remove-prop-types",
+    {
+      "removeImport": true,
+      "additionalLibraries": [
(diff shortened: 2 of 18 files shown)


nolanlawson:
On second thought, I would really like to solve this in such a way that we still retain the prop-types for developers. Otherwise there is little point in having them at all.

nolanlawson:
Hm, I can't seem to get `babel-plugin-transform-react-remove-proptypes` to work properly in `"wrap"` mode, even when downgrading due to https://github.com/oliviertassinari/babel-plugin-transform-react-remove-prop-types/issues/172. (When downgraded, it just prepends a `false ? ` check which is not what we want.)

Also even if I could get it to work, potentially it would add its own issues since we'd have introduced `process` globals into our `dist` code where it didn't exist before.

> we could make these code changes but _not_ actually run the transform ourselves, but then provide instructions in the readme for configuring the transform. 

^ This seems like the most sensible and least invasive option.

nolanlawson:
OK, now this PR is just a refactoring to make prop-types removable, plus some instructions in the README for how to remove prop-types.

nolanlawson:
This seems pretty uncontroversial to me now, so I'm going to self-approve and merge

PR #260, opened 2018-12-18
feat: add slim builds for modern browsers

### LINKED ISSUE #259: Slimmer build for modern browsers
For those of us who don't have to support IE, it would be nice to have a build of `emoji-mart` that removes all the `babel-runtime` polyfills, class transforms, etc. Perhaps the build structure could be something like:

```
dist
dist-es
dist-modern
```

…where `dist-modern` is also an ES module build, and doesn't contain any of the Babel transforms for older browsers. If users want to get fancy with this (e.g. support certain versions of certain modern browsers), then they can run their own Babel transforms.

nolanlawson:
fixes #259

This adds `dist-modern` as proposed in that issue. It also adds some text to the `README` so that folks know how to use the "slim" build.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,6 +3,21 @@
   "plugins": [
     "check-es2015-constants",
     "transform-object-rest-spread",
+    "transform-runtime",
+    [
+      "module-resolver",
+      {
+        "alias": {
+          "babel-runtime/core-js/object/get-prototype-of": "./src/polyfills/objectGetPrototypeOf",
+          "babel-runtime/helpers/extends": "./src/polyfills/extends",
+          "babel-runtime/helpers/inherits": "./src/polyfills/inherits",
+          "babel-runtime/helpers/createClass": "./src/polyfills/createClass",
+          "babel-runtime/helpers/possibleConstructorReturn": "./src/polyfills/possibleConstructorReturn",
+          "babel-runtime/helpers/classCallCheck": "./src/polyfills/classCallCheck",
+          "babel-runtime/core-js/object/keys": "./src/polyfills/keys"
+        }
+      }
+    ],
     [
       "transform-define", "scripts/define.js"
     ]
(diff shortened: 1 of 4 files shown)


nolanlawson:
For the record, in my own app this reduces the total cost of the `nimble-picker` chunk from 46.17kB to 36.04kB (22% reduction), when combined with #257.

screenshot from 2018-12-18 14-00-34

nolanlawson:
This PR is complicated by #258; I'd like to solve #258 first.

nolanlawson:
I scaled back the ambition of #258 (https://github.com/missive/emoji-mart/pull/258#issuecomment-470776501), so now I think this PR stands on its own. Assuming the library consumer has their Babel plugins set up correctly, they can use `/dist-modern` plus `babel-plugin-transform-react-remove-prop-types` to get the smallest possible production build.

nolanlawson:
Rebased and fixed merge conflicts

nolanlawson:
I'm starting to wonder if it would make more sense to release a new version of emoji-mart with a breaking change, which is to drop support for old browsers (unless you apply Babel transforms yourself to emoji-mart, in which case they are supported).

EtienneLem:
I was wondering the same yesterday. Not too sure what’s the current state of affairs in regard to component libraries. Wouldn’t it be ideal to just provide the most modern version of it and people can transpile it with their already-in-place compiling tool of choice. I guess these usually ignore `/node_modules` libraries? Or not even?

nolanlawson:
@EtienneLem Yeah and we are also already encouraging people to do this to remove the prop-types. How about this:

- Release `dist-modern` for now
- In the next major release, make modern the default

BTW by "modern" I mean "evergreen browsers, no IE11" :slightly_smiling_face:. I could make that clearer in the docs.

nolanlawson:
Rebased and tweaked the README a bit

nolanlawson:
Managed to get the size down by tweaking the `.babelrc`. The resulting on-disk sizes are:

```
204K    dist
204K    dist-es
196K    dist-modern
```

nolanlawson:
Just tested this out in my own project, and using the `remove-prop-types` Babel transform plus this branch, the `nimble-picker` is down to 35.36kB.
Screenshot from 2019-03-09 09-33-11

laurensclaessen:
@nolanlawson By moving all these `"transform-es2015-arrow-functions"` values away from the default set of plugins. It seems the `legacy-cjs` build doesn't make use anymore of this legacy set of plugins.

`legacy-cjs` is the default `dist`, but is not compatible anymore with IE11. There are `const`s and arrow functions in the build. This doesn't really match the README.md statement:

> `dist` is the standard build with the highest level of compatibility.

I noticed your discussion related to evergreen browsers, but it would be nice if this version 2.x could still remain compatible with IE11.

Thanks!

cc @EtienneLem

Issue #252, opened 2018-12-05
Split emoji rendering


adambowles:
Emoji Mart is rendering emojis that are a sequence of other emojis on multiple lines (in the case of this screenshot ['male student' is 'man' + 'graduation cap'](https://emojipedia.org/male-student/)).

Screenshot taken on macOS 10.14.1, Firefox 63.0.3.

![image](https://user-images.githubusercontent.com/10128574/49510401-fa610f00-f87f-11e8-94a7-756b802e6f53.png)

adambowles:
Safari 12.0.1 on macOS 10.14.1 is rendering the component emojis:
![image](https://user-images.githubusercontent.com/10128574/49511606-83c61080-f883-11e8-9eb9-f7b6c1bc0d2d.png)

AaronGoodrum:
Seeing the same thing on Win 10 pro, Firefox 64.
![firefox](https://user-images.githubusercontent.com/30958265/50702154-efc3a200-1004-11e9-9f6a-836fab97812c.jpg)

nolanlawson:
fixed in https://github.com/missive/emoji-mart/pull/261

PR #261, opened 2019-01-17
fix: forces emoji's not to break line

vs-ryann:
closes issue [#252 ](https://github.com/missive/emoji-mart/issues/252)

On Firefox and Safari, emojis were larger than their 24px container and were splitting into multiple rows causing formatting issues. Adding this css rule will force the emoji's to not break.


nolanlawson:
LGTM, thanks!

PR #268, opened 2019-02-05
fix: fix preact compat

### LINKED ISSUE #254: Preact compatibility: error on undefined parent in memoizeSize()
Hi there, great library, thanks so much for making it. :slightly_smiling_face: 

I'm working on creating an emoji picker that lazy-loads `emoji-mart` as well as React (since my project does not use React). I've found [remount](https://github.com/rstacruz/remount) to be a great way to bundle this library as a custom element, but I'd also like to use Preact to get the bundle size down. 

Unfortunately when I use Preact with the standard Webpack configuration:

```js
resolve: {
  alias: {
    'react': 'preact-compat',
    'react-dom': 'preact-compat'
  }
}
```

This results in the following error in the console:

```
category.js:84 Uncaught (in promise) TypeError: Cannot read property 'getBoundingClientRect' of undefined
    at Category.memoizeSize (category.js:84)
    at NimblePicker.updateCategoriesSize (nimble-picker.js:436)
    at NimblePicker.componentDidUpdate (nimble-picker.js:187)
    at renderComponent (preact.mjs:601)
    at rerender (preact.mjs:88)
```

Here's where the error happens:
screenshot from 2018-12-17 19-14-38

Interestingly, it doesn't seem to actually affect the functionality of the component itself, but it is a noisy error in the console. Maybe this is a bug in `preact-compat`, or maybe this is fixable in `emoji-mart`?

nolanlawson:
fixes #254

I'm not sure why Preact is subtly different from React, but not caching `this.parent` fixes the error. I don't think there's any performance hit for _not_ caching it (`container.parentElement` should be a cheap operation), so it seems worth it to get compatibility with Preact.

You can test this by using [this branch](https://github.com/nolanlawson/emoji-mart/tree/fix-preact-compat-test) and running the storybook. Before this fix, there's an error in the console; after this fix, there isn't.

### DIFF
--- a/src/components/category.js
+++ b/src/components/category.js
@@ -16,8 +16,6 @@ export default class Category extends React.Component {
   }
 
   componentDidMount() {
-    this.parent = this.container.parentNode
-
     this.margin = 0
     this.minMargin = 0
 
@@ -67,10 +65,10 @@ export default class Category extends React.Component {
 
   memoizeSize() {
     var { top, height } = this.container.getBoundingClientRect()
-    var { top: parentTop } = this.parent.getBoundingClientRect()
+    var { top: parentTop } = this.container.parentNode.getBoundingClientRect()
     var { height: labelHeight } = this.label.getBoundingClientRect()
 
-    this.top = top - parentTop + this.parent.scrollTop
+    this.top = top - parentTop + this.container.parentNode.scrollTop
 
     if (height == 0) {
       this.maxMargin = 0


EtienneLem:
Nice find, thanks. Sorry for the long delay on PRs, I’ll be reviewing / merging them very soon.

nolanlawson:
No problem. BTW I just did a bit of extended testing on this, and I think we're still not preact-compatible. There seem to be some bugs with switching categories. Maybe it's best to just put preact compatibility on the back burner, since both inferno and react-lite work.

EtienneLem:
Roger that. I’m currently working on a Preact project myself, so I don’t mind giving it a go once the other PRs have been sorted out.

Issue #210, opened 2018-08-03
Support passing custom className


goodmind:
Add forwarding of `className` prop to root div on `Picker` component

nolanlawson:
See discussion in https://github.com/missive/emoji-mart/pull/274#issuecomment-470969125, this feature is not necessary because you can style using selectors based on parent elements.

goodmind:
@nolanlawson but how about using with styled-components?

nolanlawson:
I don't know much about styled-components. Is there something about their approach that is incompatible with using CSS selectors?

goodmind:
@nolanlawson, no but passing `className` is more convenient
```js
const StyledPicker = styled(Picker)`
   background-color: #000;
`
```
would pass generated `className` to `Picker`

Boooober:
Also, styled-wrappers leads to unnecessary DOM nodes. In case when there are too many, for example `Emoji` components, that could be an issue.

Using them in this style would be quite handy:
```javascript
import styled from 'styled-components';
import { Emoji as EmojiMart } from 'emoji-mart';

export const Emoji = styled(EmojiMart)`
  vertical-align: middle;
  // any other styles
`;
```

PR #274, opened 2019-03-01
Add support for passing a custom class name to all components

HenriBeck:
This adds support for passing a `className` prop to all of the exposed components.

Fixes: #210

### DIFF
--- a/src/utils/shared-props.js
+++ b/src/utils/shared-props.js
@@ -24,6 +24,7 @@ const EmojiPropTypes = {
   ]),
   size: PropTypes.number.isRequired,
   emoji: PropTypes.oneOfType([PropTypes.string, PropTypes.object]).isRequired,
+  className: PropTypes.string,
 }
 
 const EmojiDefaultProps = {
@@ -37,9 +38,6 @@ const EmojiDefaultProps = {
   tooltip: false,
   backgroundImageFn: (set, sheetSize) =>
     `https://unpkg.com/emoji-datasource-${set}@${EMOJI_DATASOURCE_VERSION}/img/${set}/sheets-256/${sheetSize}.png`,
-  onOver: () => {},
-  onLeave: () => {},
-  onClick: () => {},
 }
 
 const PickerPropTypes = {
@@ -79,12 +77,11 @@ const PickerPropTypes = {
   notFound: PropTypes.func,
   notFoundEmoji: PropTypes.string,
   icons: PropTypes.object,
+  className: PropTypes.string,
+  emojiClassName: PropTypes.string,
 }
 
 const PickerDefaultProps = {
-  onClick: () => {},
-  onSelect: () => {},
-  onSkinChange: () => {},
   emojiSize: 24,
   perLine: 9,
   i18n: {},
@@ -105,7 +102,6 @@ const PickerDefaultProps = {
   autoFocus: false,
   custom: [],
   skinEmoji: '',
-  notFound: () => {},
   notFoundEmoji: 'sleuth_or_spy',
   icons: {},
 }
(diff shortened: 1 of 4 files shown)


nolanlawson:
Just for context, what problem does this solve? If you'd like to style the emoji-picker, you can always use a class selector that relies on a parent (or some other technique). For instance:

```html
<div class="parent">
  <div class="emoji-mart">
    <!-- etc -->
  </div>
</div>
```

```css
.parent .emoji-mart {
  /* etc */
}
```

 Is there a case where you need to set a class on the emoji-picker itself?

HenriBeck:
Just wrapping the picker around another DOM node for styling the picker is in my opinion bad practice.

Also, this could potentially break some layouts because of the added node.

src/utils/shared-props.js
```diff
@@ -37,9 +38,6 @@ const EmojiDefaultProps = {
   tooltip: false,
   backgroundImageFn: (set, sheetSize) =>
     `https://unpkg.com/emoji-datasource-${set}@${EMOJI_DATASOURCE_VERSION}/img/${set}/sheets-256/${sheetSize}.png`,
-  onOver: () => {},
-  onLeave: () => {},
-  onClick: () => {},
```
nolanlawson: Why remove these onOver/onLeave/onClicks?

src/utils/shared-props.js
```diff
@@ -105,7 +102,6 @@ const PickerDefaultProps = {
   autoFocus: false,
   custom: [],
   skinEmoji: '',
-  notFound: () => {},
```
nolanlawson: Ditto

EtienneLem:
> Just wrapping the picker around another DOM node for styling the picker is in my opinion bad practice.

I’m curious why that would be. It’s arguably exactly what cascading means 😄

> Also, this could potentially break some layouts because of the added node.

This will not break emoji-mart layout. As for the layout around emoji-mart, I’m not sure I see how adding emoji-mart doesn’t break the layout, but having emoji-mart inside let’s say another `<div>` would 🤔

If I’m being honest, I feel like this is a little bit superfluous. (It also doesn’t add new className to all the existing `emoji-mart-*` classes)

nolanlawson:
On second thought, yeah I agree with @EtienneLem. It's not like you have to add a second `<div>` - there must be _some_ element wrapping `emoji-mart`, which you can use to style it. If it's the only `emoji-mart` on your page, then it's even easier - just add your own CSS to target the `.emoji-mart` classes.

Plus adding this to the codebase means one more thing to maintain and document... I would argue for resolving this as wontfix. Thanks very much for the contribution though; it highlights the fact that perhaps we could do a better job documenting how to style emoji-mart.

topolanekmartin:
> 
> 
> > Just wrapping the picker around another DOM node for styling the picker is in my opinion bad practice.
> 
> I’m curious why that would be. It’s arguably exactly what cascading means 😄
> 
> > Also, this could potentially break some layouts because of the added node.
> 
> This will not break emoji-mart layout. As for the layout around emoji-mart, I’m not sure I see how adding emoji-mart doesn’t break the layout, but having emoji-mart inside let’s say another `<div>` would 🤔
> 
> If I’m being honest, I feel like this is a little bit superfluous. (It also doesn’t add new className to all the existing `emoji-mart-*` classes)

I am sorry but I don't agree with your opinion mess up DOM with another node only for styling your component. It is not good for many reasons. What is so wrong with adding property className to the component's root element?

EtienneLem:
> […] mess up DOM with another node only for styling your component.

You don’t have to. You can target `.emoji-mart` from your own stylesheets to customize 100% of EmojiMart styles.

PR #277, opened 2019-03-08
Support case where showPreview is false but showSkinTones is true. Also add prepare script.

kayleemann:
It is useful to be able to disable the mouseover emoji preview area but still have showSkinTones available. In particular, in order to be able to use the component on iOS without having to double-click every time it is necessary to disable much mouseover behavior such as this. (see https://css-tricks.com/annoying-mobile-double-tap-link-issue/)

Also, it is useful to have a "prepare" script in the package.json so that it is possible to clone emoji-mart directly from a git repo without extra steps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,8 @@
     "prepublishOnly": "npm run build",
     "storybook": "start-storybook -p 6006",
     "build-storybook": "build-storybook",
-    "prettier": "prettier --write \"{src,scripts,spec}/**/*.js\""
+    "prettier": "prettier --write \"{src,scripts,spec}/**/*.js\"",
+    "prepare": "npm run build:dist"
   },
   "size-limit": [
     {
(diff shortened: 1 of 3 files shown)


nolanlawson:
Could you provide some more details about what this PR is doing? I'm not sure I follow what exactly this fixes. Presumably mouseover is a desktop-only convention, but I suppose this change is mostly for mobile?

Also the `"prepare"` thing seems like a great fix but unrelated, would be good to have it in a separate PR since it seems very uncontroversial to me. :)

nolanlawson:
Nevermind, after reading through the CSS Tricks article and this PR again, I understand. You're not trying to solve the iOS hover issue per se - that's just what led you to want to hide the mouseover preview but show the skin tones. However that doesn't work, so you opened this PR. Got it.

LGTM, merging!

PR #278, opened 2019-03-08
fix: fix babel-runtime not found

### LINKED ISSUE #228: Can't resolve babel-runtime - doesn't work with Babel v7
Building a project that includes `emoji-mart` but not `babel-runtime` fails:

```
ERROR in ./node_modules/emoji-mart/dist-es/utils/index.js
Module not found: Error: Can't resolve 'babel-runtime/core-js/object/keys' in '[...]/node_modules/emoji-mart/dist-es/utils'

ERROR in ./node_modules/emoji-mart/dist-es/components/skins.js
Module not found: Error: Can't resolve 'babel-runtime/helpers/classCallCheck' in '[...]/node_modules/emoji-mart/dist-es/components'

...etc.
```

I see in `.babelrc` that some Babel imports are aliased to polyfills by the `module-resolver` plugin, but not all of them, so the dist files depend on packages that aren't necessarily installed.

For example, this can happen if a project is using Babel v7, which has different package names.

Perhaps the project should explicitly include Babel in `dependencies` instead of `devDependencies`, or else make sure all the imports are covered by polyfills?

nolanlawson:
fixes #228

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -34,7 +34,9 @@
           "babel-runtime/helpers/extends": "./src/polyfills/extends",
           "babel-runtime/helpers/inherits": "./src/polyfills/inherits",
           "babel-runtime/helpers/createClass": "./src/polyfills/createClass",
-          "babel-runtime/helpers/possibleConstructorReturn": "./src/polyfills/possibleConstructorReturn"
+          "babel-runtime/helpers/possibleConstructorReturn": "./src/polyfills/possibleConstructorReturn",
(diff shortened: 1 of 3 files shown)


nolanlawson:
This PR ensures that we are not shipping `babel-runtime`, so we truly have 0 dependencies. It does so by inlining some missing polyfills.

nolanlawson:
This fix will unblock a lot of people who are on Babel 7, and it's pretty minor, so merging now.

PR #282, opened 2019-03-09
fix: make categories into accessible nav buttons

### LINKED ISSUE #219: A11Y: Category Icons should be tabbable
The icons on top of the picker should be reachable in the tab order. There may also be an appropriate aria role for a nav menu or something similar that could be used

nolanlawson:
fixes #219

1. Changes the categories `<div>` into a `<nav>` (see [navigation role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Navigation_Role) for details).
2. Changes each category button into a true `<button>`. This makes it tabbable, and you can press the <kbd>space</kbd> button to press them.
3. Adds `aria-label` for each button, since the `title` is not sufficient for a11y
4. Adds an `aria-label` to the `<nav>` to help distinguish it from any other `<nav>`s that may be on the page. This requires a new `i18n` label ("Emoji categories" by default).

Here is a GIF of me pressing <kbd>tab</kbd> and <kbd>spacebar</kbd> to demonstrate: https://gfycat.com/agedwindycockatiel

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ import { Picker } from 'emoji-mart'
 search: 'Search',
 notfound: 'No Emoji Found',
 skintext: 'Choose your default skin tone',
+categorieslabel: 'Emoji categories', // Accessible title for the list of categories
 categories: {
   search: 'Search Results',
   recent: 'Frequently Used',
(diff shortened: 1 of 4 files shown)


nolanlawson:
Fixed a small CSS issue - I noticed that the user agent stylesheet for `<button>`s can override the `fill` now, so we can't use `currentColor`.

PR #286, opened 2019-03-10
perf: cache the scrollbar width

nolanlawson:
When loading the storybook, `measureScrollbar()` is called 15 times. Each time, we insert and remove elements from the DOM, and invoke style/layout. This seems wasteful, since the width of a scrollbar is unlikely to change during the lifecycle of the page. We should be able to just measure it once and cache the measured value.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -171,8 +171,13 @@ function deepMerge(a, b) {
   return o
 }
 
+let scrollbarWidth
+
 // https://github.com/sonicdoe/measure-scrollbar
 function measureScrollbar() {
+  if (typeof scrollbarWidth === 'number') {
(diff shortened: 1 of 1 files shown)


nolanlawson:
On simulated 4x slowdown in Chrome on my Dell XPS 13 I clocked `measureScrollbar()` as 56ms total (during page load of the storybook, so 15 calls). Not huge, but enough to make this a win.

nolanlawson:
On second thought, this may introduce bugs because the scrollbar can change width (e.g. if custom scrollbar CSS changes), and it's not really a huge win. Closing for now

nolanlawson:
Whoops, didn't mean to reopen

PR #290, opened 2019-03-10
fix: fix search input when value is prepopulated

nolanlawson:
In some cases (notably with Preact it seems), the `<input type="search">` value may be prepopulated (due to cached DOM nodes being re-hydrated, I suppose). In those cases, you can cause a bug if you do e.g.:

1. Type "horse" into the search
2. Press enter to choose the :horse: emoji
3. Reopen the picker (can't be reproduced in the storybook, but can be reproduced in [Pinafore](http://github.com/nolanlawson/pinafore/))
4. The search input is prepopulated, but the search results are empty

This fixes that, by performing a search if the input happens to contain any text.

### DIFF
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -26,6 +26,13 @@ export default class Search extends React.PureComponent {
     this.handleChange = throttleIdleTask(this.handleChange.bind(this))
   }
 
+  componentDidMount() {
+    // in some cases (e.g. preact) the input may already be pre-populated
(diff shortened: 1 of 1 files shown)


nolanlawson:
Interestingly, this bug seems fixed in `preact@10.0.0-alpha.1` (the search input is not pre-populated), but since it's a small amount of code, it seems worthwhile to do what we can to avoid an outright bug.

PR #295, opened 2019-03-12
fix: use tabIndex, not tabindex

nolanlawson:
This is something I missed in #283. Apparently React uses `tabIndex`, not `tabindex`.


nolanlawson:
Self-merging without waiting for review because this is a pretty obvious bug

Issue #294, opened 2019-03-12
Grab-bag of accessibility issues


nolanlawson:
After merging #281 #282 #283 #284, there are still some remaining a11y issues I'd like to tackle:

- [x] title emoji is decorative, should be aria-hidden
- [x] title text should be a title for the whole thing, using `aria-labelledby` or something
- [x] skintone picker should better communicate pressed and dropdown states
- [x] the list of emoji in each category should use `<ul>`/`<li>`
- [x] overall structure of regions/landmarks could be improved

nolanlawson:
There could be additional improvements, especially in the skintone picker, but for now I think these are largely fixed.

PR #297, opened 2019-03-13
fix: improve a11y of skin tone picker

nolanlawson:
I don't think this is a full fix for #294 but it's a bit better than before. When the picker is not expanded, it's a popup button, then when you click it it becomes a toggle button within a menu item within a menu of similar buttons.

I tested this in NVDA and VoiceOver, and it works much better in NVDA than VoiceOver (VoiceOver doesn't seem to notice the toggle state for the buttons, nor the menu). But it's usable in either one.


nolanlawson:
Apparently Preact doesn't like the syntax `{...(condition && {foo: 'bar'})}`, so I'm using `{...(condition ? {foo: 'bar'} : {})}` instead.

PR #308, opened 2019-03-18
keep short_names on sanitize

hinaloe:
https://github.com/missive/emoji-mart/blob/d739f27714c4b1e28f3d92f654a5ba83933c2816/src/components/emoji/nimble-emoji.js#L118

throws `TypeError: Cannot read property '0' of undefined`. (It makes https://github.com/tootsuite/mastodon/issues/10315 )

This is one of way to resolve issue, but it may not a good way.

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -31,6 +31,7 @@ function sanitize(emoji) {
     return {
       id,
       name,
+      short_names,
       colons,
       emoticons,
       custom,


nolanlawson:
Thank you!

nolanlawson:
And yes we could probably go through the codebase and try to handle all cases where `short_names` is undefined, but this solution seems simpler.

nolanlawson:
Wait, don't we need `short_names` here as well? 

https://github.com/hinaloe/emoji-mart/blob/a1a12abcd5751ca20597afd3eb787812d1262efa/src/utils/index.js#L46-L54

hinaloe:
Oh, I didn't face the problem with non-custom emoji, but it looks to need that to make interface uniform!

nolanlawson:
I also can't seem to write a Jest test to reproduce this, but I am new to Jest.

PR #315, opened 2019-03-22
test: add basic skins test

nolanlawson:
Trying to improve the test infrastructure of emoji-mart. I want to test actual behavior, but ideally without launching an actual browser (or JSDom). Seems Enzyme is one of the most popular ways to do this with Jest. So this adds a simple test for the skintone picker.


nolanlawson:
More tests are always preferable to fewer tests, merging

PR #317, opened 2019-03-22
fix: consistently return short_names from sanitize

nolanlawson:
Based on #308 I am just going to assume that we always need to return `short_names` here. Admittedly, though, I haven't managed to write a Jest test to actually reproduce #308, nor can I prove that this is actually currently causing a bug. But it seems harmless to return `short_names` here.

### DIFF
--- a/src/utils/emoji-index/__tests__/emoji-index.test.js
+++ b/src/utils/emoji-index/__tests__/emoji-index.test.js
@@ -5,6 +5,7 @@ test('should work', () => {
     {
       id: 'pineapple',
       name: 'Pineapple',
+      short_names: ['pineapple'],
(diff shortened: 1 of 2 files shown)


nolanlawson:
@hinaloe fyi

hinaloe:
Looks good!

I reproduced the issue with jest, but something is not good enough
https://gist.github.com/hinaloe/a9a5f1cdc3d7134466eaf27e784ffd05

nolanlawson:
Awesome, thank you. I've turned your diff file into a PR here and given you credit as the author: https://github.com/missive/emoji-mart/pull/323

I ran your tests and they definitely reproduce the issue! So that's wonderful. Thanks for your help. :)

PR #321, opened 2019-03-23
Return skin emoji on search

pederjohnsen:
Returns emojis with skin data using skin tone from store in searches.

This is useful when doing emoji search for an autocomplete component.
Previously it would just return the default emoji data for an emoji, but now we get the skin related data too depending on what skin tone is set in the store.

Split out from #250


src/utils/emoji-index/nimble-emoji-index.js
```diff
@@ -33,7 +35,16 @@ export default class NimbleEmojiIndex {
         })
       }
 
-      this.emojis[id] = getSanitizedData(id, null, null, this.data)
+      // If skin variations include them
+      if (skin_variations) {
+        this.emojis[id] = {}
+        for (let skinTone = 1; skinTone <= 6; skinTone++) {
+          this.emojis[id][skinTone] = getSanitizedData({id, skin: skinTone}, skinTone, this.set, this.data)
+        }
```
nolanlawson: This could probably be smaller if it were an array, but an object of IDs 1-6 is fine

nolanlawson:
Awesome PR. The only small issue is that we need to run `yarn prettier` to apply the right code style, but other than that this is great. Love the new tests as well. Tested it out in the storybook and it's definitely an improvement. Thanks!

pederjohnsen:
@nolanlawson hmm that's odd, I have prettier setup in my editor to format on save and looks like there is a prettier config in the repo, so would've thought it should've caught any issues :thinking:

PR #329, opened 2019-03-28
fix: use span for non-clickable emojis

### LINKED ISSUE #309: Add new <Emoji/> component or prop that doesn't use <Button/>?
I use the `<Emoji/>` component in multiple places where it doesn't need to be clickable, seems odd to render it as a `<Button/>` in those cases.

Might be nice to either add a new component that renders it visually as before, or add a prop to switch between button and non button?

EtienneLem:
Fixes #309

- No more default `onClick` handler (undefined `props.onClick` was already handled)
- When `onClick` is defined, use `<Button>` with `type="button"` (see https://github.com/missive/emoji-mart/issues/309#issuecomment-476087144)


nolanlawson:
Awesome, great choice. :+1:

PR #332, opened 2019-04-05
changed style in emoji-mart-category-list. added display css.

cygnetmaulikp:
image

style is broken when i created storybook in my project.

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -163,6 +163,7 @@
 .emoji-mart-category-list {
   margin: 0;
   padding: 0;
+  display: block;
 }
 
 .emoji-mart-category-list li {


EtienneLem:
While this is a very simple change, it feels like the issue might come from your own global styles in your project. Any chance you’re setting `ul` default `display` to something else? `emoji-mart-category-list` is a `ul` which should already be `display: block` by default, unless overwritten in your CSS.

Again, I know this is very little bytes added but emoji-mart probably already being the heaviest dependency in term of weight, I’d rather not include fixes in the upstream that are not present by default.

Adding `.emoji-mart-category-list { display: block }` to your project would be the best way to solve this. We have a lot of such fixes in our private codebase as well 😄

EtienneLem:

PR #338, opened 2019-05-01
Added `type={"button"}` to anchors.js

The-Code-Monkey:
without this it will submit the form if you are within a redux form with react not sure about other libraries but the button doesn't seem to need to not have this type attribute

### DIFF
--- a/src/components/anchors.js
+++ b/src/components/anchors.js
@@ -43,6 +43,7 @@ export default class Anchors extends React.PureComponent {
               aria-label={i18n.categories[id]}
               title={i18n.categories[id]}
               data-index={i}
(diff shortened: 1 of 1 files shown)


The-Code-Monkey:
i can confirm this fixes the issue and doesn't break any functionailty

The-Code-Monkey:
@missive can we merge this please really affecting a project that im currently working on and cannot edit node_modules every time i add something

EtienneLem:
Yes of course, sorry for the delay and thanks for the patch!

EtienneLem:
`v2.11.1` has just been released.

PR #343, opened 2019-05-14
Deterministic sorting of search results.

xaviershay:
This is important for reliable tests, e.g. if you are snapshotting views
and comparing to reference images.

Pair with https://github.com/thinkproductivity/emoji-mart-native/pull/62


nolanlawson:
There's a small style fix (prettier), after that LGTM!

PR #352, opened 2019-06-29
fix: ensure unique keys for React list

### LINKED ISSUE #327: Search bug
When you write "fam" in search field you see duplicate key error in console.
Снимок экрана 2019-03-26 в 12 58 52

If you continue write "family" you will see more same emoji.

nolanlawson:
fixes #327

Probably the `id` in the index itself should be guaranteed to be unique as well, but this at least fixes the React warning.


nolanlawson:
time-bombing this PR after4 days

PR #358, opened 2019-08-20
Keep auto focus when changing category

pederjohnsen:
Suggestion:

If autofocus is true, keep search input focused when changing between emoji categories to still allow for quick search.

### DIFF
--- a/src/components/picker/nimble-picker.js
+++ b/src/components/picker/nimble-picker.js
@@ -388,6 +388,8 @@ export default class NimblePicker extends React.PureComponent {
     } else {
       scrollToComponent()
     }
+
+    this.search.focus()
   }
 
   handleSkinChange(skin) {
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -21,6 +21,7 @@ export default class Search extends React.PureComponent {
     this.setRef = this.setRef.bind(this)
     this.clear = this.clear.bind(this)
     this.handleKeyUp = this.handleKeyUp.bind(this)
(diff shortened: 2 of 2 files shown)


krvajal:
Any reason for not merging this?

pederjohnsen:
@krvajal don't think any PRs have been merged since July, hopefully this will be looked at eventually and merged in :crossed_fingers: :pray:

nolanlawson:
The current docs say `autoFocus` means "Auto focus the search input when mounted." Could it be a bit confusing for some people to add this behavior?

I also worry about the accessibility implications – having the focus jump to the search box may be unexpected.

pederjohnsen:
@nolanlawson most other emoji pickers I've seen in the wild keeps focus on the search input on category change, and is the reason I suggested this change, but you have some valid points there :ok_hand:

nolanlawson:
I can see some alternative approaches:

- Make it a new option (`autoFocusOnSectionChange`)
- Fire an event when the section changes and let users do whatever they want (implicitly, you can focus by doing `document.querySelector('.emoji-mart input[type=search]').focus()`).

Personally I think `autoFocus` should maybe be removed entirely, since 1) users can do it themselves as described above, and 2) in my experience it doesn't work consistently.

EtienneLem:

I think it would make sense, Apple’s emoji picker seems to behave this way. We do have to reinstate a way to programmatically focus the search input in v5.0.1, I’ll consider automatically refocusing the input when clicking on a category.

PR #362, opened 2019-09-09
Ensure emoji backgrounds do not tile

### LINKED ISSUE #361: Non-square emoji are misaligned and repeat
The emoji `<span>` uses `background-image`, but doesn't set `background-repeat` or `background-position`. This means that non-square emoji will tile, and align to the left rather than the centre.

This occurs in the list and in the detail view:
Screen Shot 2019-09-09 at 16 14 44
Screen Shot 2019-09-09 at 16 14 35

rmccue:
Fixes #361.


nolanlawson:
Cool, thanks!

PR #368, opened 2019-10-02
Feat: Adding functionality to parse emoji from text

Dipen-Dedania:
- Ability to lookup emoji within the text.


nolanlawson:
Thanks for the suggestion! Is there any reason this couldn't be a separate package? It seems like the logic is mostly self-contained, and everything it needs from `emoji-mart` is exposed as a public API.

The only reason I say this is that it would be one more thing for `emoji-mart` to maintain...

EtienneLem:

Issue #356, opened 2019-08-13
We should modify Nimble-Picker for supporting React 17.x


h6ak:
`componentWillReceiveProps` will be removed in React 17.x.
We should modify Nimble-Picker component for supporting React 17.x.

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

viiiprock:
Vote !

MateuszDev96:
I'm waiting too :(

nolanlawson:
Seems fixed by #372, please reopen if I'm wrong

PR #372, opened 2019-10-21
Use getDerivedStateFromProps instead of componentWillReceiveProps

h6ak:
`componentWillReceiveProps` will be removed from React 17.x. So, I use `getDerivedStateFromProps` instead of it. 

This pull request solves issue #356.


nolanlawson:
Thanks!

Issue #293, opened 2019-03-12
Add dark mode


nolanlawson:
It may be nice to have an optional dark mode.

darwin2k:
you can try to overwrite CSS:
```
.emoji-mart, .emoji-mart-category-label span {
  @include themify($themes) {
    background-color: themed('colorBackground');
    color: themed('colorText');
  }
}
```

nolanlawson:
fixed by https://github.com/missive/emoji-mart/pull/391

santoshnaagula:
you can add darkmode true or false, it will reflect

[darkMode]="false"

beqacrc:
at the end you can use this solution

```scss
@use "../../abstract/palette.scss" as *;

@mixin emoji-mart-on-dark {
  .emoji-mart {
    background-color: $clr-main-primary;

    .emoji-mart-preview-name {
      color: #ccc;
    }

    .emoji-mart-category-label {
      background-color: unset;
    }

    .emoji-mart-category .emoji-mart-emoji:hover:before,
    .emoji-mart-emoji-selected:before {
      background-color: $clr-blue-primary;
    }
  }
}


```

PR #374, opened 2019-10-26
ISSUE #293 Enhancement: Add Dark Mode

zolomohan:
## Dark Mode

image

This can be attained by passing a prop to Picker Component.

`import { Picker } from 'emoji-mart'`
...
`<Picker darkMode={true}>`

The Default Prop is false.

This PR Fixes Issue #293 . Enhancement: Add Dark Mode

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -160,11 +160,13 @@
   top: 0;
 }
 
-.emoji-mart-category-label span {
+.emoji-mart-category-label-span {
   display: block;
   width: 100%;
   font-weight: 500;
   padding: 5px 6px;
+  background-color: #fff;
+  color: #000;
 }
 
 .emoji-mart-category-list {
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -4,6 +4,11 @@
   line-height: 1.15;
 }
 
+.emoji-mart-dark-bg{
+  color: #fff !important;
+  background-color: #222 !important;
+}
+
 .emoji-mart {
   font-family: -apple-system, BlinkMacSystemFont, "Helvetica Neue", sans-serif;
   font-size: 16px;
@@ -143,6 +148,10 @@
   border-radius: 100%;
 }
 
+.emoji-mart-category .emoji-mart-emoji-dark:hover:before{
+  background-color: #444;
+}
+
 .emoji-mart-category-label {
   z-index: 2;
   position: relative;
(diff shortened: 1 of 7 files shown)


css/emoji-mart.css
```diff
@@ -4,6 +4,11 @@
   line-height: 1.15;
 }
 
+.emoji-mart-dark-bg{
+  color: #fff !important;
+  background-color: #222 !important;
+}
```
nolanlawson: Using `!important` here feels a bit hacky. Can we be more targeted with this?

css/emoji-mart.css
```diff
@@ -151,13 +160,17 @@
...
   font-weight: 500;
   padding: 5px 6px;
-  background-color: #fff;
-  background-color: rgba(255, 255, 255, .95);
+  background-color: rgb(255, 255, 255, 0.95);
+  color: #000;
+}
+
+.emoji-mart-category-label-span-dark {
+  background-color: rgba(0, 0, 0, 0.95) !important;
```
nolanlawson: Ditto about `!important` here.

src/utils/shared-default-props.js
```diff
@@ -31,6 +31,7 @@ const PickerDefaultProps = {
   emojisToShowFilter: null,
   showPreview: true,
   showSkinTones: true,
+  darkMode: false,
```
nolanlawson: The default should probably be based on prefers-color-scheme: https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

nolanlawson:
I also think a simpler solution may be to just add the `emoji-mart-dark` CSS to the top level, and then use CSS to select for that (rather than add a new class to all the children). For instance:

```css
.emoji-mart-dark .something-inside-of-emoji-mart {
  background: black;
}
```

nolanlawson:
I made the changes suggested in my PR review and pushed it above ^. I only tweaked the design slightly (changed the background color of the sections and added a white fill to the search icon):

Screenshot from 2019-12-21 09-09-39

nolanlawson:
I'm going to merge as two separate PRs. Thanks very much for this!

nolanlawson:
Closing in favor of https://github.com/missive/emoji-mart/pull/391

zolomohan:
> I'm going to merge as two separate PRs. Thanks very much for this!

You're very welcome!

PR #377, opened 2019-11-03
Move `With custom data` to correct position

pederjohnsen:
Looks like the `With custom data` section of `Headless search` got pushed down by the `Get emoji data from Native` stuff.


nolanlawson:
Thanks!

PR #385, opened 2019-12-07
fix: gracefully ignore bad JSON in localstorage

DrFrankenstein:
# Background
We were previously using an [Angular port](https://github.com/TypeCtrl/ngx-emoji-mart) of emoji-mart, then we migrated our application to React, and began using the original emoji-mart.  That port used the `emoji-mart.last` localstorage item slightly differently, by not JSON-encoding it. After the migration, our app would crash while rendering emoji-mart, as it attempted to parse the stale value from our users' localstorage.

# Changes
This PR changes the store's behaviour to discard invalid JSON values instead of raising an error through `JSON.parse`. That way, if those values ever become corrupt or otherwise invalid for some reason, things like recent emoji just get forgotten, instead of the app inexplicably crashing.

### DIFF
--- a/src/utils/store.js
+++ b/src/utils/store.js
@@ -44,12 +44,12 @@ function get(key) {
     if (!isLocalStorageSupported) return
     try {
       var value = window.localStorage[`${NAMESPACE}.${key}`]
+
+      if (value) {
+        return JSON.parse(value)
+      }
     } catch (e) {
       return
-    }
-
-    if (value) {
-      return JSON.parse(value)
     }
   }
 }


src/utils/store.js
```diff
@@ -44,13 +44,13 @@ function get(key) {
     if (!isLocalStorageSupported) return
     try {
       var value = window.localStorage[`${NAMESPACE}.${key}`]
+
+      if (value) {
+        return JSON.parse(value)
```
tvanier: Could/should it be `_JSON` like in the [set](https://github.com/missive/emoji-mart/blob/master/src/utils/store.js#L35) function?

tvanier: Could/should it be `_JSON` like in the [set](https://github.com/missive/emoji-mart/blob/master/src/utils/store.js#L35) function?

src/utils/store.js
```diff
@@ -44,13 +44,13 @@ function get(key) {
...
+      }
     } catch (e) {
       return
     }
-
-    if (value) {
-      return JSON.parse(value)
-    }
```
tvanier: Catch as much as possible?
```js
function get(key) {
  try {
    if (getter) {
      return getter(key)
    }

    if (!isLocalStorageSupported) return
    
    var value = window.localStorage[`${NAMESPACE}.${key}`]
    if (value) {
      return _JSON.parse(value)
    }
  } catch (e) {
    return
  }
}
```

src/utils/store.js
nolanlawson: I think that `_JSON` should be removed entirely; not sure why it's there.

src/utils/store.js
nolanlawson: This seems like a good idea given that on some environments (e.g. iOS Safari in "no cookies" mode) localStorage throws an error if you try to access it.

nolanlawson:
LGTM! Thanks

PR #386, opened 2019-12-18
Fix: Allowing emoji selection on enter

### LINKED ISSUE #353: Cannot select emoji via 'Enter' on keydown
When Picker is open, you can tab through the emojis but you cannot select one via the enter key. The onSelect property is never recognized when selecting this way.

abrioux:
The `handled` variable should only be set when we actually call the `handleEmojiSelect` function, otherwise it overrides default button behaviour of being selectable by the enter key. 

This causes an issue for accessible keyboard navigation: https://webaim.org/techniques/keyboard/

Fixes #353


nolanlawson:
Thanks!

PR #394, opened 2019-12-24
fix: remove core-js dependency

### LINKED ISSUE #393: Implicit dependency on core-js
Unfortunately, it looks like we allowed `core-js` to sneak into our dependencies, due to `Object.values()` (#351) and `JSON.stringify` (#392).

This adds about ~5kB extra to the bundle size, and of course will break if someone doesn't have `core-js` in their dependencies, since it's an implicit dependency.

nolanlawson:
Fixes #393 

This removes `core-js` from our output bundle... for now. In the future, we need to come up with a better system for ensuring that Babel doesn't include `core-js` in our bundle.

### DIFF
--- a/src/components/picker/nimble-picker.js
+++ b/src/components/picker/nimble-picker.js
@@ -105,7 +105,9 @@ export default class NimblePicker extends React.PureComponent {
         this.CUSTOM.push(customEmoji)
       })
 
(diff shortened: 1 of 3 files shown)


nolanlawson:
It's been 6 days, merging

PR #400, opened 2019-12-29
test: test bundle size in CI

nolanlawson:
We have this test (to keep the size under 80kB), but it's not currently running in CI. This fixes that.


nolanlawson:
It's been 2 weeks and this is pretty trivial, merging

PR #401, opened 2019-12-29
fix: remove rAF polyfill, add SSR test

nolanlawson:
As far as I can tell, the rAF polyfill is only needed for SSR support (#7). [rAF has been available in browsers since IE10](https://caniuse.com/#feat=requestanimationframe) so it's unlikely we need it for browser support.

This PR adds a test (`yarn test:ssr`) to make sure that SSR is working correctly, and also removes the rAF polyfill. As the test demonstrates, we don't actually use rAF in any of the `render()` function (only on button clicks or similar interactions), so we can safely remove it AFAICT.


nolanlawson:
Rebased, fixed conflicts

Issue #366, opened 2019-09-27
Disable sorting of frequently used emojis on click


vrizo:
Hi, @EtienneLem !

Thanks for Emoji Mart — an excellent project which we use on our project!

We faced a problem reported by our users. Sorting of frequently used emojis is annoying because it fires when the user clicks on the emoji.

<img src="https://media.amplifr.com/images/social_images/013/286/618/img.gif" alt="Catch me if you can" width="402" height="412" />

What do you think about disabling sorting on the fly in favor of sorting after component mounting?

I found the same issue here, but the author decided to close it: https://github.com/missive/emoji-mart/issues/265. I don't think that it is a bug, but I think it might be improved.

Thank you!

vrizo:
Hey, @EtienneLem @nolanlawson !

Do you have any thoughts on this?

I noticed that there is the same behavior on macOS in the native mini picker. But it works in full-size native picker in another way. 1:1. Attaching a GIF.

<img src="https://media.amplifr.com/images/social_images/013/477/642/img.gif" alt="macOS emoji picker" />

Thank you!

vrizo:
Hi, @EtienneLem @nolanlawson ! Do you have any ideas about this feature? Thanks!

edgesinghkataria:
We can give a property to enable/disable this functionality.
Let me know what suits best.
I've fixed this In my current project, will be happy to submit a PR

EtienneLem:
I feel like this should be the default and be re-sorted only when mounting

nolanlawson:
I agree, having the icons move out from under your mouse is confusing, even if the native macOS emoji picker also does this.

SaraRandolph:
Hi all! I have been working on a solution for this and I'm wondering if you guys would be open to reviewing a PR for this issue. This will be my first PR for an open source project so it won't be perfect but it will at least be on the road to a solution for this enhancement!

EtienneLem:
@SaraRandolph Absolutely! You will notice that we’ve been pretty slow on reviewing/implementing lately. I will have some time to allocate on this very soon ✌️

SaraRandolph:
@EtienneLem sounds good! no rush! I'll make the pr this afternoon. thank you!

EtienneLem:
Fixed with #417

PR #417, opened 2020-03-04
enhancement: functionality to disable the frequent emoji sorting

SaraRandolph:
This is in reference to issue #366.

I added:

- a bool value `disableFrequentEmojiSort` that defaults to false 
- functionality in storybook so we can all see it in action

I'm very open to any feedback anyone has so please let me know if I missed anything...its been known to happen 😉😂

Thank you guys!

### DIFF
--- a/src/components/picker/nimble-picker.js
+++ b/src/components/picker/nimble-picker.js
@@ -289,7 +289,7 @@ export default class NimblePicker extends React.PureComponent {
     var component = this.categoryRefs['category-1']
     if (component) {
       let maxMargin = component.maxMargin
-      if (!this.props.disableFrequentEmojiSort){
+      if (!this.props.disableFrequentEmojiSort) {
         component.forceUpdate()
       }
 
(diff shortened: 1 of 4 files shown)


EtienneLem:
Oh, it was that easy huh? 😅 

One might argue that it shouldn’t even be an option, but the default non-configurable behaviour. What do you think? 🤔It does feel weird and hardly expected that these can change position when you want to insert multiple of the same emoji.

SaraRandolph:
@EtienneLem hmmm I know what you are saying about it seeming like unexpected behavior when clicking on it. BUT maybe since thats how its been built people who use emoji-mart are excepting that behavior and any change would be upsetting? 🤷

If it were just up to me, I would say we make the default behavior be to NOT update the emojis and give people to option to configure it, if they so choose. I can adjust the PR to reflect that and change the bool value to `enableFrequentEmojiSort`, if we want. 

However, these things are not up to just me 😂 so I think that with either option, we can easily justify the decision. 🎉

EtienneLem:
> If it were just up to me, I would say we make the default behavior be to NOT update the emojis and give people to option to configure it, if they so choose. I can adjust the PR to reflect that and change the bool value to `enableFrequentEmojiSort`, if we want.

I did just that, I’m confident this is the expected behavior by most. Will add this as a breaking change in the changelog.

Thanks for contributing @SaraRandolph 🙌

Simon-Laux:
wouldn't be `disableImmediateFrequentEmojiSort` a better name?

Issue #360, opened 2019-09-04
Cannot read property 'value' of null


minhtranite:
I use Sentry to log errors and there is an issue with emoji-mart:

![Screen Shot 2019-09-04 at 5 31 01 PM](https://user-images.githubusercontent.com/5418628/64265501-3a633880-cf5d-11e9-8277-70ba69aa5054.png)

nolanlawson:
Hm, I'm not sure how this could possibly happy. It can only happen if the `input` (`this.input`) doesn't exist, but the search somehow happens. Do you have a test to reproduce? Are you using some React-alike (such as Preact or Inferno)?

theopak:
reported as closed by #419

EtienneLem:
Chances are this is fixed in latest version. Please reopen an issue if that’s not the case.

PR #419, opened 2020-03-18
safely access this.input.value

mattkrick:
not really sure how to reproduce, but apparently one of my users managed to time their search perfectly (probably due to the throttle?).

this should prevent that.

Screen Shot 2020-03-18 at 11 15 10 AM

### DIFF
--- a/src/components/search.js
+++ b/src/components/search.js
@@ -65,7 +65,9 @@ export default class Search extends React.PureComponent {
   }
 
   handleChange() {
-    this.search(this.input.value)
+    if (this.input) {
+      this.search(this.input.value)
+    }
   }
 
   handleKeyUp(e) {


h6ak:
This PR will resolve https://github.com/missive/emoji-mart/issues/360 .

CoryDanielson:
This fix will work (thank you!)

An alternative to this is to update `throttleIdleTask` to return a cancel method on the returned function that you can call on `componentWillUnmount`. This is available when using throttle and debounce in lodash so it's sort of a common convention with debounced/throttled stuff.

https://lodash.com/docs/4.17.15#throttle
> Creates a throttled function that only invokes func at most once per every wait milliseconds. The throttled function comes with a cancel method to cancel delayed func invocations and a flush method to immediately invoke them.

Usage:
```js
constructor(props) {
  // ...
  this.handleChange = throttleIdleTask(this.handleChange.bind(this))
}

componentWillUnmount() {
  this.handleChange.cancel();
}
```

This would take more effort but is generally a cleaner fix because it allows you to write your callbacks without having to handle the scenario where a component may be unmounted.


/2 cents from someone who discovered this error in Sentry today

felipevegaaraujo:
Can we get this fix merged/released?

EtienneLem:
Thanks 🤘🏼, sorry for the delay. **v3.0.1** has just been released.

PR #422, opened 2020-04-24
added padding around emoji hover

anatolzak:
Currently, when hovering over the emoji, the emoji is not centered inside the round hover area.

image

I added padding to the ::before and now when hovering over an emoji, it is perfectly centered. 🙂

image

### DIFF
--- a/css/emoji-mart.css
+++ b/css/emoji-mart.css
@@ -144,6 +144,7 @@
   width: 100%; height: 100%;
   background-color: #f4f4f4;
   border-radius: 100%;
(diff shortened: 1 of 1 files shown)


EtienneLem:
This is specific to Windows native emojis and is a bit more complicated to fix than adding more padding to all emojis. This would break all non-native sets and native Apple emojis:

Screen Shot 2020-04-24 at 10 44 47 AM

Windows emojis are so big, we would need to detect when on a Windows machine and when the native set is used, reduce the `font-size`.

Issue #348, opened 2019-06-24
Better keyboard navigation/selection


rugk:
## Use case

You have `autoFocus` on and now you want to search for an emoji and quickly select it (preferably without a mouse, because that's faster).

## Currently

What you can do currently is:
* search for it
* press <kbd>Tab</kbd> multiple times to navigate to the right position (two times as far as I see) to get to the first emoji
* use <kbd>Tab</kbd> again (or <kbd>shift</kbd>+<kbd>Tab</kbd>) to go through the list item by item…
* enter/space to select emoji

## Proposed solution
* When you are in the input field an <kbd>🔽</kbd> (arrow down) should bring you (the focus) to the list of emojis, i.e. actually to the first emoji in that list.
* The focus style should be improved, so you can _see_ what (also an accessibility improvement) is selected.
* use the keyboard keys to navigate the list of emojis in a grid-style (**currently**: if you use up/down arrows, it scrolls the scrollbar)
* Just press <kbd>space</kbd> or <kbd>enter</kbd> as usual to select an emoji.
* When you are at the top row of your emojis. <kbd>🔼</kbd> (arrow up) should bring you (the focus) to the search bar.
* <kbd>Tab</kbd> navigation should obviously still be possible.

nolanlawson:
These are great ideas, although I would add that there should be an `aria-describedby` that explains for the benefit of screenreader users how to use these shortcuts. I'm not sure how to surface these shortcuts to non-screenreader users though.

rugk:
Okay, good idea. However, I'd argue that:
1. This feature is **not only** for screenreaders. Actually, I would like to use it by myself and I don't use a screenreader. (And it has been suggested in my project by a user who apparently also does not use a screenreader: https://github.com/rugk/awesome-emoji-picker/issues/41).
2. I personally think a description is not really necessary (for users that don't use a screenreader, at least), because that is quite intuitive IMHO. I tried pressing the down button in the search input field, because that's the direction where the emojis are.
  And navigating the list of emojis with arrow keys is something very intuitive, because that's how any game navigation or so works… :wink: 

So anyway yes, we can add those.

jonahparampara:
I agree with rugk, and I think the main benefit are for general powerusers who are good with their keyboards, not screenreaders.

There also seems to be a bug in the current selection implementation - if I use `tab` to navigate to a search results, and then hit `enter`, the first search result is returned instead of the selected item.

xindixu:
[...]

PR #431, opened 2020-05-22
feat/add keyboard control

xindixu:
# Changes
closes TypeCtrl/ngx-emoji-mart#313 
Added keyboard control:

## Category anchors:
ezgif com-video-to-gif (1)

- When focusing on any emoji, <kbd>Tab</kbd> to change focus to the first category anchor
- When a category is selected, <kbd>Arrow Down</kbd> to go to the first item in the category 
- This also means that when focusing on any emoji, <kbd>Tab</kbd> can no longer used to go to the next emoji
- Same applies to search
ezgif com-video-to-gif (2)

## Emoji:
ezgif com-video-to-gif

- <kbd>Arrow Up</kbd>, <kbd>Arrow Down</kbd>, <kbd>Arrow Left</kbd>, <kbd>Arrow Right</kbd> to go to emojis on top/down/left/right to the current emoji
- Able to jump to next/previous category if at the last line/first row

## Style:
- emoji focus style now should look the same as emoji on hover style

## Markup:
- Change emoji layout to table/cell with a11y roles

References:
- This PR is largely inspired by #348. Note that some of the features mentioned are not implemented or implemented differently. 
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Grid_Role
- Slack emoji picker


rugk:
Awesome PR! Thanks! :tada: 
Really glad, I want to see this merged.

> This PR is largely inspired by #348. Note that some of the features mentioned are not implemented or implemented differently.

Uhm, care to say which exactly?

The far I see these changes in your very nice gifs, I am very satisfied. The only thing I possibly miss is the focus style:
> The focus style should be improved, so you can see what (also an accessibility improvement) is selected.

rugk:
Does this possibly also fix https://github.com/TypeCtrl/ngx-emoji-mart/issues/313?

Also note you can (automatically) [let issues close](https://help.github.com/articles/closing-issues-via-commit-messages/) when a PR is merged by adding some "magic" text to your PR _body_.

xindixu:
This PR differs from #348 in these ways:
 
> - [x] When you are in the input field an 🔽 (arrow down) should bring you (the focus) to the list of emojis, i.e. actually to the first emoji in that list.
> - [ ] The focus style should be improved, so you can see what (also an accessibility improvement) is selected.
> - [x] use the keyboard keys to navigate the list of emojis in a grid-style (currently: if you use up/down arrows, it scrolls the scrollbar)
> - [x] Just press space or enter as usual to select an emoji.
> - [ ] When you are at the top row of your emojis. 🔼 (arrow up) should bring you (the focus) to the search bar.
> - ~~Tab navigation should obviously still be possible.~~

- <kbd>Tab</kbd> can no longer used to select emojis. It will bring focus to the first category anchor. Users can use <kbd>Tab</kbd> to tab through category anchors and search bar. Users can also use arrow keys to select emojis

xindixu:
> The focus style should be improved, so you can see what (also an accessibility improvement) is selected.
@rugk I am not sure the best way to improve this. I just used the same mouse hover style it has currently. Note that developers using this library can still manually override focus style.

xindixu:
> Does this possibly also fix [TypeCtrl/ngx-emoji-mart#313](https://github.com/TypeCtrl/ngx-emoji-mart/issues/313)?
Yup, it will. I'll add magic to close this issue. As for #348, since not all of feature you mentioned is implemented, should I still close the issue?

rugk:
> Tab can no longer used to select emojis. It will bring focus to the first category anchor.

I'm not sure whether this is a good idea, accessibility-wise. As that is the default accessibility feature. But I'm probably the wrong one to ask, whether this is good…

> As for #348, since not all of features you mentioned is implemented, should I still close the issue?

I'd say yes, as long as the maintainer(s) do not say anything else.

rugk:
>  When you are at the top row of your emojis. arrow_up_small (arrow up) should bring you (the focus) to the search bar.

Even if you are at the first of _all_ rows? E.g. this should be possible IMHO (+ user story):
1. Search for `wink`
2. Go down -> Only now notice you want to use a different emoji
3. Press up again -> "_Correct_" your search term to `upside` [down face] e.g.

xindixu:
> > When you are at the top row of your emojis. arrow_up_small (arrow up) should bring you (the focus) to the search bar.
> 
> Even if you are at the first of _all_ rows? E.g. this should be possible IMHO (+ user story):

I'd say that since emoji grid is separated from the search bar, based on the a11y guideline, <kbd>Up</kbd> should not go back to search bar since it is essentially different region. Maybe we could have search bar always in focus and thus users could type in more characters if they need to.
So it will be:
1. Search for wink
2. Go down -> Only now notice you want to use a different emoji
3. Hit backspace and directly type in 'upside'

haji-ali:
If I may, I would like to offer some feedback on the interface.

I think having the search bar always in focus makes more sense to me since it’s unambiguous. Letters and backspace change the search field, arrows select a particular emoji, Enter selects the emoji. 
Selecting categories can be done with Tab and Shift-Tab. Clearly this makes navigating the text field impossible but since the search text is expected to be short I don’t think that should be an issue. 

If the search bar is auto-focused when the interface is shown, that would allow selecting an emoji using the keyboard only with minimal keystrokes, which is ideal while writing a message.

xindixu:
All great ideas. I'm going to implement 
- search bar always in focus and user can type anywhere to search 
- auto select first emoji in the category when a category is selected or search result when search is triggered

rugk:
> search bar always in focus and user can type anywhere to search

Does not this conflict with accessibility though? Don't screenreaders always read the thing that is focused?

src/components/picker/nimble-picker.js
```diff
@@ -278,10 +284,161 @@ export default class NimblePicker extends React.PureComponent {
...
+      return lastEmojiIndex
+    }
+
+    let newRow
+    let newColumn
+    let newCategoryIndex = categoryIndex
+    let emojiIndex
+    const lastEmojiIndex = getLastEmojiIndex(categoryIndex)
+
+    switch (e.keyCode) {
```
rugk, line 315: > This feature is no longer recommended.
> […]
> You should avoid using this if possible; it's been deprecated for some time. Instead, you should use KeyboardEvent.code, if it's implemented. Unfortunately, some browsers still don't have it, so you'll have to be careful to make sure you use one which is supported on all target browsers.

https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Could we try not introducing legacy code here, possibly?

xindixu:
Could we consider merging this PR in?

src/components/picker/nimble-picker.js
```diff
@@ -439,8 +596,8 @@ export default class NimblePicker extends React.PureComponent {
   handleKeyDown(e) {
     let handled = false
 
-    switch (e.keyCode) {
-      case 13:
+    switch (e.key) {
+      case 'Enter':
```
rugk, line 600: Just because I'm not sure: Currently (with <kbd>tab</kbd> focus switches) also <kbd>space</kbd> works for selecting emojis. This here only refers to <kbd>Enter</kbd>.
IMHO <kbd>space</kbd> should also continue to work… :thinking:

rugk:
I also don't know whether this is the right thing to add in this PR, but [I have a use case](https://github.com/rugk/awesome-emoji-picker/issues/100), where I'd like more shortcuts to trigger something in the picker/with the selected emoji. In my case Ctrl+C should e.g. trigger a copy to clipboard.

As all methods we can use to get the selected/chosen emoji currently are _reactive_ (i.e. as an dev outside the lib you only get the emoji if something happens as as a result of a click or so) this is hard to implement.
A _active_ method that could just return the currently focused emoji could be very helpful for that. (Note this use case barely existed before, because if you click an emoji you get the reaction.)
So e.g. a `getCurrentlyFocussedEmoji` that likely is used internally in a similar way, may be very useful here.

xindixu:
@rugk Sorry for the late reply. What you mentioned is a pretty cool idea. I think you could create a separate issue and PR on this. :)

rugk:
Okay, done: https://github.com/missive/emoji-mart/issues/444

animify:
Looking forward to this too, very much needed for full keyboard navigation 💯

xindixu:
If you are interested in adding keyboard support, feel free to check out this repo: https://github.com/pingboard/emoji-mart.

castroCrea:
Hey @EtienneLem any idea when you can merge that and release ?
Thanks 
And thanks @xindixu and @rugk

hackel:
Please consider also including vi-style keyboard navigation (hjkl) in addition to the arrow keys.  This would make navigation even more efficient while typing, not having to move one's hands from the home row at all.  I don't believe this would conflict with anything.

AjayKumar-25:
Any tentative date of merging and releasing this ?

EtienneLem:
Sorry for the radio silence on this PR.

There are some important changes that I’m not very sold on, I’m not sure why the layout had to be changed to `<table>`. I’d be favorable for a more accessibility-based implementation where we can use `onFocus` and move the focus with keyboard shortcuts.

xindixu:
@EtienneLem The main reason why I changed to `table` is that: `table` is a markup more suitable with the aria-role `grid`. 
`role=grid` implies exisitance of <kbd>Arrow Up</kbd>, <kbd>Arrow Down</kbd>, <kbd>Arrow Left</kbd>, <kbd>Arrow Right</kbd> navigations. See more: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Grid_Role

evang522:
I would really love to see this merged in and part of a newer version. I'm planning to use this for a Chrome extension I'm developing, but in my opinion it's really not that user friendly unless keyboard navigation is possible.

rugk:
> I'm planning to use this for a Chrome extension I'm developing, but in my opinion it's really not that user friendly unless keyboard navigation is possible.

I totally agree and I'm actually affected in the same way as you are affected.
Because I already have a Firefox extension based on this library, which could also be ported/made compatible with Chrome/ium browsers. This is tracked here: https://github.com/rugk/awesome-emoji-picker/issues/54
It was just not on my highest priority so far.

EtienneLem:

The new version has full keyboard control.

PR #445, opened 2020-07-30
Option to provide an own class name to custom emojis

ErikMichelson:
Sometimes you need to set some special CSS attributes to make a custom emoji perfectly displayed. Until now, you only could style all custom emojis in general with the `emoji-mart-emoji-custom` class. But if you use emoji images from different sources, you maybe want to change some settings of individual custom emojis.

This PR adds support for an attribute `className` which may be added to the declaration of a custom emoji. The value of this attribute will then be added as class name to the custom emoji element.


ccoenen:
This looks interesting! What's the status on this?

EtienneLem:

PR #449, opened 2020-08-29
Add possibility to add window context from which requestAnimationFrame should be fired

dominiczaq:
When using incorrect window context some browsers (e.g. firefox) will silently fail to call requestAnimationFrame.

This PR introduces windowContext prop which allows to pass window object directly into `NimblePicker` component.


EtienneLem:

PR #453, opened 2020-10-08
Allow providing an Emojiindex instead of always creating a new one

heikomat:
Because of the `JSON.parse(JSON.stringify(...))` in `getData()` the creation of the NimbleEmojiIndex is rather slow.
This change allows the user to provide an already created EmojiIndex to the picker, so it doesn't need to be recreated every time a new picker instance is created.

In a use-case i have this saves about 30-40ms of render-time every time a popup with a picker is opened.


5ebastianMeier (approved):
LGTM

webnugget (approved):

EtienneLem:

PR #479, opened 2021-01-18
remove inappropriate emoji

kushagra-saxena:


EtienneLem:
Any change to data would need to be done on https://github.com/iamcal/emoji-data

PR #480, opened 2021-01-20
Fix aria-label and title for custom categories

ehinson:
The aria-label and title weren't showing up for the Custom button.  

Screen Shot 2021-01-20 at 3 59 59 PM


EtienneLem:
Thanks 🤘🏼, sorry for the delay. **v3.0.1** has just been released.

Issue #447, opened 2020-08-11
Support for relative sizes on the Emoji component


BrockWills:
First of all, thank you for this incredible project.

I was hoping to pass a responsive size such as `size="2rem"` to the `<Emoji />` component.

With the current codebase this seems to work perfectly fine because the size is just passed straight through to css, but PropTypes validation fails because it expects a number. What do people think about updating the PropTypes for the Emoji component to instead be:
```
size: PropTypes.oneOfType([
  PropTypes.string,
  PropTypes.number
]).isRequired,
```

That way, all types of css sizes like `px`, `em`, `rem`, etc. would all be supported.

vple:
I'd like this as well and just made a PR.

I messed around with it a bit—this change will support rem as expected, but it doesn't work as intended with em and %. Details below.

The Emoji component renders something like this:

```
<span aria-label="👍, +1, thumbsup" class="emoji-mart-emoji emoji-mart-emoji-native">
  <span style="font-size: 10rem;">👍</span>
</span>
```

The issue for em is that `.emoji-mart-emoji` sets `font-size` to 0. So any em value for the inner span will also be 0. That was introduced [here](https://github.com/missive/emoji-mart/commit/919976e10e9f42187aeb925767910d731ff2252e), but I'm not familiar enough to know if there are good alternate approaches.

The issue for % is that the outer span doesn't have an explicit size, so the inner span gets confused.

EtienneLem:
This is now possible on latest version with the `<em-emoji>` web component. See https://github.com/missive/emoji-mart#-emoji-component

PR #507, opened 2021-06-07
Allow string emoji sizes

vple:
This allows for other sizes (1rem, 1em, 10%) in addition to the current numerical sizes, which are treated as px.

Although this does support em and %, note that they don't render as expected:
- The .emoji-mart-emoji class sets font-size to 0, which results in any em value resolving to 0.
- % based sizes will be weird, since the parent span doesn't have an explicit size.

See also: https://github.com/missive/emoji-mart/issues/447


EtienneLem:

PR #510, opened 2021-06-22
Add virtual list support using react-window

sayjeyhi:
This PR is adding virtual list support.

all features seem to be working. but the sticky category label will be missed


yannterzi:
@EtienneLem si tu as un peu de temps, cette PR est pas mal importante pour réduire l'impact de performance sur nos projets

smatt989:
why hasn't this been merged yet?

hoangvu12:
Works great, but still got some error like this

image

sayjeyhi:
@hoangvu12 I don't have that bug, which OS and browser you are facing that bug? also can you give a repo to reproduce that issue?

hoangvu12:
> @hoangvu12 I don't have that bug, which OS and browser you are facing that bug? also can you give a repo to reproduce that issue?

I'm using Opera GX and Windows 10. Sorry I don't have any repo to reproduce it.

hoangvu12:
image
This is how it looks like in Microsoft Edge

image
I noticed that the `li` tag is empty. Maybe it is failed to get html?

sayjeyhi:
Well, I think you are using native emoji, and some emojis are not visible there, so a quick solution could be using those emoji sprites, like `apple`, `google`, or other emoji sets.

hoangvu12:
> Well, I think you are using native emoji, and some emojis are not visible there, so a quick solution could be using those emoji sprites, like `apple`, `google`, or other emoji sets.

No I'm currently using `facebook` set.

sayjeyhi:
so can you check why that specific emojis are not loading? check with CSS background

hoangvu12:
> so can you check why that specific emojis are not loading? check with CSS background

It doesn't has any child element that use CSS background
image

sayjeyhi:
well this seems to be another issue, and I think it has nothing to do with the virtual scroll 🤔

hoangvu12:
It only happens in that emoji category, I haven't tried another set yet and other emojis working just fine.
Also, I have a live demo, do you want to check it out?

EtienneLem:
Thanks a lot for your contribution. [v5.0.0](https://github.com/missive/emoji-mart/discussions/573) has been released and this doesn’t apply anymore. The new version should handle an infinite (give or take) amount of emojis.

Issue #416, opened 2020-03-04
Export `ref` and `searchRef`


jonahparampara:
The picker seems to have a `ref` prop but it references the react component instead of the DOM component as expected.

Furthermore it would be really nice to export the search input ref as a `searchRef` prop.

In my case I want this so I can focus the search input on command (`autoFocus` doesn't suffice my use-case and I need more control)

jsheffers:
I'm specifically interested in forwarding the ref mentioned above so that my useRef returns the DOM element instead of the react component.  I'm unable to do a ref.current.contains(e.target) because the ref itself returns the component.  This would give me the ability to see the user clicked inside of the picker.  Is there another recommended way to do this with functional components?

jsheffers:
FYI @jonahparampara.  I was able to achieve what I wanted by creating a simple wrapping div and applying the ref to it.  Not sure if that helps in your case.

Rafi993:
Hi I would like to have `searchRef` too. If this issue is not worked I can help with this :)

arclogos132:
Any update on this?

tommoor:
Thanks @jantonso – this is reasonable workaround. setTimeout shouldn't be needed with `useLayoutEffect`, but depends on your situation.

PR #516, opened 2021-07-29
setSearchRef function

Rafi993:
## Why?

Sometimes you might want to access ref of search input in the parent component if autoFocus is not enough and you need more control. Right now that prop is not possible

## Related issues

closes https://github.com/missive/emoji-mart/issues/416

## How does this work?

1. This PR adds a prop `setSearchRef` which would allow you to pass a function which will set the search ref 
2. The reason it did not use ref forwarding is because if we want to add more refs for other things this keeps things simple and also this method of passing function to set ref is also used internally 

### Does this code have tests ?
- I might need some help writing tests for this

### README / Documentation update ?
- Yes I have updated README.md with prop in the prop list

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ import { Picker } from 'emoji-mart'
 | **notFoundEmoji** | | `sleuth_or_spy` | The emoji shown when there are no search results |
 | **notFound** | | | [Not Found](#not-found) |
 | **icons** | | `{}` | [Custom icons](#custom-icons) |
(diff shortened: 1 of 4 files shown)


EtienneLem:
Thanks a lot for your contribution. While [v5.0.0](https://github.com/missive/emoji-mart/discussions/573) has been released and this doesn’t apply anymore, a similar API will most likely be needed.

Something to think about for v5.0.1, thanks again!

PR #539, opened 2021-10-25
Add lazy load to custom emojis

### LINKED ISSUE #492: emoji picker slow to render to DOM
would be nice to have a virtualized list so that we don't need to wait for the hundreds of emojis to render, which can really slow down the browser.  are there any plans for this (or some other way to speed up the rendering)?

mashirozx:
fix #492


EtienneLem:

PR #562, opened 2022-03-26
Allow emoji to have a class given

test137E29B:
A small update in which I have added the ability to pass through a Class Name that gets merged with the other class names in the Nimble Emoji component.

This feature would be a great addition, as it allows some repositioning and movement of the final emoji, which may be required when using external packages such as MUI. A problem I encountered using this component is that the Emoji were always ~4px above center line, when using centered emoji.

Example Use Case:
_before_
image

_after_
image


EtienneLem:

PR #566, opened 2022-04-04
feat: add support for React.js v18

### LINKED ISSUE #565: Support React.js v18
Hey :wave:,

Now that React.js v18 has been released, it could be cool to support it.

Ref: https://github.com/missive/emoji-mart/blob/ccd591dd5af5ea21455e65b16c233f4876447b80/package.json#L27

Divlo:
Fixes #565 

I don't know how you test different React versions in your CI but it might be worth running tests with React.js v18.
it might also be worth removing support for some older React.js versions, so it is easier to maintain, is `emoji-mart` really support React.js v0.14.0?


Divlo:
cc @EtienneLem

PR #583, opened 2022-05-03
add translation brazilian portuguese

arthur-es:


EtienneLem:
🇵🇹 🇧🇷 Obrigada!

Quick question: On the public website, would `Português` be the correct spelling in the locale select?

PR #589, opened 2022-05-07
Add italian translation

starise:


EtienneLem:
🇮🇹 Grazie!

Quick question: On the public website, would `Italiano` be the correct spelling in the locale select?

starise:
Yes, `Italiano` is correct. :+1:

PR #592, opened 2022-05-09
Add Chinese translations

akirarika:
This is a Chinese translations.

I use sense-for-sense translation as much as possible to ensure the use experience of Chinese users 😀


EtienneLem:
🇨🇳 谢谢！

Quick question: On the public website, would `中文` be the correct spelling in the locale select?

PR #578, opened 2022-04-27
Add German translation

bennetfabian:

PR #596, opened 2022-05-12
Corrected DE translation location

bennetfabian:
Corrects a mistake I accidentally made in #578


EtienneLem:
Thanks! I missed it as well in the original PR

PR #641, opened 2022-07-18
Fix custom element not reading `skin` attribute

srmagura:


EtienneLem:
This has been handled in db1725f

PR #642, opened 2022-07-18
Fix `init` function being a no-op if no arguments were passed

srmagura:
Note that the new test will print this error: https://github.com/missive/emoji-mart/issues/646. It doesn't actually cause any problems though, it's just an annoyance.


EtienneLem:
Not needed since db1725f

PR #650, opened 2022-07-22
Allow to use an id for custom categories

Clebal:
When I add a custom emoji category, I cannot indicate a custom id, so a default one is used what is a bit uncomfortable when I have to fill the categories array.

Current case:

```
init({
    ...,
    custom: [{ name: "Basic", emojis: [ … ] }],
    categories: ["custom_1"]
})
```

After this change:

```
init({
    ...,
    custom: [{ id: "basic", name: "Basic", emojis: [ … ] }],
    categories: ["basic"]
})
```

### DIFF
--- a/packages/emoji-mart/src/config.js
+++ b/packages/emoji-mart/src/config.js
@@ -66,7 +66,9 @@ async function _init(props) {
 
       if (!category.emojis || !category.emojis.length) continue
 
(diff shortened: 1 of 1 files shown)


EtienneLem (approved):

EtienneLem:
Totally makes sense, thank you!

PR #661, opened 2022-08-04
Add burger as hamburger keyword

Clebal:
Currently, when you search "burger" you get nothing:

image

I think it makes sense to add `burger` as one of the hamburger keywords.

Another alternative would be to make search more flexible, but that can lead to major changes in search results.


EtienneLem:
Thanks for the PR!

JSON files aren’t meant to be edited manually, they are generated by a script. This kind of change (keywords) would either need to be handled in this file (https://github.com/missive/emoji-mart/blob/main/packages/emoji-mart-data/build.js) or within the lib we use for keywords (https://github.com/muan/emojilib).

This however made me realize that **burger** should’ve already been a keyword and fixed an issue in 8977a6d. It should now match in the latest [@emoji-mart/data](https://www.npmjs.com/package/@emoji-mart/data) release.

PR #662, opened 2022-08-05
Add Czech localization

potty:


packages/emoji-mart-website/index.html
```diff
@@ -64,6 +64,7 @@
           ><select id="locale" class="flex-grow">
             <option value="en">English</option>
             <option value="ar">العربية</option>
+            <option value="cs">Czech</option>
```
EtienneLem, line 67: In this list we display the language in its native form. According to my search, it would be **čeština** for Czech, is that accurate?

EtienneLem (changes requested):

packages/emoji-mart-website/index.html
potty: I'm sorry I overlooked that, you are right, **Čeština** is the correct word :) I'll send fix

EtienneLem (approved):

EtienneLem:
🇨🇿 Dík!

PR #666, opened 2022-08-10
Make selected emoji accessible to screen readers

Charca:
## Problem

Users can navigate emojis using the arrow keys on their keyboards. Unfortunately, this behavior does not announce to screen reader users what the currently selected emoji is.

## Solution

Add an [ARIA live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) only visible to screen readers (hidden from the UI) that announces the currently selected emoji.

## Demo

This can be tested using a screen reader like VoiceOver on the Mac 👇

CleanShot 2022-08-10 at 14 12 28

### DIFF
--- a/packages/emoji-mart/src/components/Picker/Picker.js
+++ b/packages/emoji-mart/src/components/Picker/Picker.js
@@ -964,6 +964,13 @@ export default class Picker extends Component {
     )
   }
 
+  renderLiveRegion() {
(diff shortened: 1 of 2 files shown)


EtienneLem (approved):

EtienneLem:
Very nice, thanks! 🤘🏼

PR #687, opened 2022-08-29
Fix border color example

santiagendrix:


santiagendrix (commented):
The css propierty `--color-border-hover` doesn't exist on the src.

santiagendrix:
Maybe is even better to find and replace all `border-over` to `border-hover`.

EtienneLem:
Thanks! 🤘🏼

PR #699, opened 2022-09-15
Fix error when window.HTMLElement is undefined

TranquilMarmot:
This can happen in some environments, for example if a `react-native` app happens to import something from `emoji-mart`.


EtienneLem (approved):

EtienneLem:
Thanks! Could you `git rebase` please? I tried doing it on my end, but I don’t have the permission.

PR #700, opened 2022-09-19
Add callback for changing skin tone

TranquilMarmot:


packages/emoji-mart/src/components/Picker/Picker.js
```diff
@@ -26,7 +26,7 @@ export default class Picker extends Component {
 
   getInitialState(props = this.props) {
     return {
-      skin: Store.get('skin') || props.skin,
```
TranquilMarmot, line 29: Flipping this since it would be nice if the skin tone was a "controlled" value. We store our user's preferred skin tone in their application settings.

EtienneLem: This won’t work because `props.skin` is always defined and defaults to `1` when not provided. We would need to add another props, something like a `getSkin` method maybe?

```js
skin: props.getSkin ? props.getSkin() : Store.get('skin') || props.skin
```

EtienneLem (changes requested):

PR #701, opened 2022-09-21
Add "part" to root element to allow overriding CSS variables

TranquilMarmot:
This allows you to do something like this to override CSS variables in `emoji-mart`:

```css
em-emoji-picker::part(root) {
    --em-rgb-background: '51, 51, 51';
}
```

Since it renders in the shadow DOM, I couldn't figure out another way to actually do these overrides.

### DIFF
--- a/packages/emoji-mart/src/components/Picker/Picker.js
+++ b/packages/emoji-mart/src/components/Picker/Picker.js
@@ -1051,6 +1051,7 @@ export default class Picker extends Component {
     return (
       <section
         id="root"
+        part="root"
         class="flex flex-column"
         style={{
           width: this.props.perLine * this.props.emojiButtonSize + (12 + 16),


EtienneLem:
This shouldn’t be required, `--em-*` variables are meant to be private. You can change background color with:

```css
em-emoji-picker {
  --rgb-background: 262, 240, 283;
}
```

TheEssem:
There are other parts of the interface that can't be styled as easily, such as the size of the skin tone picker and search bar. This change would have allowed for somewhat easier styling of elements inaccessible via the variables, like in pre-5.0 versions.

EtienneLem:
Yes, it’s true for other parts. A PR would be welcome.

gordokan:
If this is the correct solution to adding `part` to the root element, can this PR be reopened or should a new one be created?

PR #705, opened 2022-09-27
Add optional ref to search input

TranquilMarmot:
We have a scenario where the search input can become unfocused, but then we have to focus it again. To do this, we need to be able to pass a ref to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ function App() {
 | **skinTonePosition** | `preview` | `preview`, `search`, `none` | The position of the skin tone selector |
 | **theme** | `auto` | `auto`, `light`, `dark` | The color theme of the picker |
 | **getSpritesheetURL** | `null` | | A function that returns the URL of the spritesheet to use for the picker. It should be compatible with the data provided. |
+| **searchInputRef** | `null` | | Optional ref to pass to the search input |
 
 ### Custom emojis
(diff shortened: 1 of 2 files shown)


EtienneLem (approved):

EtienneLem:
Great idea! Needs to be rebased, thanks!

README.md
```diff
@@ -115,6 +115,7 @@ function App() {
 | **skinTonePosition** | `preview` | `preview`, `search`, `none` | The position of the skin tone selector |
 | **theme** | `auto` | `auto`, `light`, `dark` | The color theme of the picker |
 | **getSpritesheetURL** | `null` | | A function that returns the URL of the spritesheet to use for the picker. It should be compatible with the data provided. |
+| **searchInputRef** | `null` | | Optional ref to pass to the search input |
```
EtienneLem, line 118: This props is missing in `PickerProps.ts`

packages/emoji-mart/src/components/Picker/Picker.js
```diff
@@ -334,7 +334,8 @@ export default class Picker extends Component {
   }
 
   handleSearchInput = async () => {
-    const input = this.refs.searchInput.current
+    const input =
+      this.props.searchInputRef?.current ?? this.refs.searchInput.current
```
EtienneLem, line 338: Would be safer to have a `this.getSearchInputRef` method to avoid having to duplicate this logic every time we need to use the search input ref.

EtienneLem (changes requested):

PR #711, opened 2022-10-05
Allow passing imageURL / spritesheetURL as specific props

### LINKED ISSUE #697: Spritesheet source is not respected for <em-emoji> element
Following on from https://github.com/missive/emoji-mart/issues/585

When displaying an emoji using `<em-emoji set="twitter" getSpritesheetURL={()=>"sprites.png"} shortcodes=":+1"></em-emoji>` the image is always fetched from the cdn.

Expected result is it uses the url provided from the function in the same way the Picker element does.

```html
<div class="flex flex-auto flex-middle flex-center" style="height: 36px; font-size: 36px;">
  <span class="emoji-mart-emoji" data-emoji-set="twitter">
    <span style="display: block; width: 36px; height: 36px; background-image: url(&quot;/emoji/sprites.png&quot;); background-size: 6100% 6100%; background-position: 93.3333% 83.3333%;"></span>
  </span>
</div>
```

Actual Result:
```html
<em-emoji spritesheet="true" set="twitter" shortcodes=":hugging_face:">
  <span class="emoji-mart-emoji" data-emoji-set="twitter">
    <img alt="珞" src="https://cdn.jsdelivr.net/npm/emoji-datasource-twitter@14.0.0/img/twitter/64/1f917.png" style="height: 1em; width: auto; display: inline-block; position: relative; top: 0.1em;">
  </span>
</em-emoji>
```

TranquilMarmot:
Fixes #697

Web components can only take in strings as props (see https://stackoverflow.com/questions/50865614/can-i-pass-function-as-attribute-to-web-component)

When you pass the `em-emoji` component a function for `getImageURL` / `getSpritesheetURL` they actually end up being `null` so this will always return `null`:

https://github.com/missive/emoji-mart/blob/2054b155cac965bb432ca6b6cd70b51304f8ed7f/packages/emoji-mart/src/config.js#L260-L264

This PR adds `imageURL` / `spritesheetURL` props that can be passed in as strings instead.

Another solution could be to export an `Emoji` component from `@emoji-mart/react` which will properly handle the function as a prop. The downside there is that the `em-emoji` web component won't really be fully usable (or it should at least be documented that it doesn't support any props that are functions)

It almost seems like the `getImageURL` / `getSpritesheetURL` props should be removed as options altogether 🤔

### DIFF
--- a/packages/emoji-mart/src/components/Emoji/EmojiProps.js
+++ b/packages/emoji-mart/src/components/Emoji/EmojiProps.js
@@ -17,6 +17,10 @@ export default {
     },
   },
 
+  imageURL: null,
+  spritesheetURL: null,
+  spritesheet: null,
+
   // Shared
   set: PickerProps.set,
   skin: PickerProps.skin,
--- a/packages/emoji-mart/src/components/Emoji/Emoji.js
+++ b/packages/emoji-mart/src/components/Emoji/Emoji.js
@@ -1,6 +1,34 @@
 import { Data } from '../../config'
 import { SearchIndex } from '../../helpers'
 
+const getImageSrc = (props, emojiSkin) => {
+  if (props.set === 'native' || props.spritesheet) {
+    return undefined
+  }
+
+  if (typeof props.getImageURL === 'function') {
+    return props.getImageURL(props.set, emojiSkin.unified)
+  }
+
+  if (typeof props.imageURL === 'string') {
+    return props.imageURL
+  }
+
+  return `https://cdn.jsdelivr.net/npm/emoji-datasource-${props.set}@14.0.0/img/${props.set}/64/${emojiSkin.unified}.png`
+}
+
+const getSpritesheetSrc = (props) => {
+  if (typeof props.getSpritesheetURL === 'function') {
+    return props.getSpritesheetURL(props.set)
+  }
+
+  if (typeof props.spritesheetURL === 'string') {
+    return props.spritesheetURL
+  }
+
+  return `https://cdn.jsdelivr.net/npm/emoji-datasource-${props.set}@14.0.0/img/${props.set}/sheets-256/64.png`
+}
+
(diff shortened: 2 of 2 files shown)


packages/emoji-mart/src/components/Emoji/EmojiProps.js
```diff
@@ -17,6 +17,10 @@ export default {
     },
   },
 
+  imageURL: null,
+  spritesheetURL: null,
```
EtienneLem, line 21: `imageURL` and `spritesheetURL` would need to be added to the README.

EtienneLem (changes requested):
Right, makes sense yes. The getters can be kept, it’s used internally when passed down from the Picker.

packages/emoji-mart/src/components/Emoji/EmojiProps.js
mageshkumar1598: @EtienneLem It seems getProps will only allow defaultProps regardless of what props we pass. I am passing `spritesheet:true` in order to use spritesheet image as source. But `getProps()` function is only generating props using defaultProps where spritesheet prop is not present!

what am I missing here!?

Could you please help? @EtienneLem

mageshkumar1598:
@EtienneLem Could you plese merge this soon?

standreinmcp:
@TranquilMarmot Hi! why was this not merged? I'm facing the same issue today. Is there another way to do this? I want to prevent fetching from CDN altogether.

TranquilMarmot:
@standreinmcp I closed it because I got tired of seeing it in my list of opened PRs after years of it not being merged, and the maintainers of this repo are MIA 😅

We ended up publishing our own version of emoji mart to our private NPM repo that included this changeset.

If I were you, I would try and find another emoji picker. I don't have any suggestions, sorry.

PR #712, opened 2022-10-07
Use maxWidth / maxHeight for image Emoji

TranquilMarmot:
We noticed that with non-square custom emoji, they would overlap each other.

By using `maxWidth` / `maxHeight` here, they will still be the right size but will not expand beyond the `size` passed in to props.


EtienneLem (approved):

EtienneLem:
Great, thanks! Would need a rebase.

EtienneLem:
Merged in 73714b3

PR #719, opened 2022-10-31
Fix Persian translation typos

masoud-msk:


EtienneLem:
@abolfazl-talaei As the original author of the Persian language file, can you confirm whether these changes make sense? According to Google Translate they would make sense, but that’s not necessarily 100% reliable so I can’t really tell myself. Thanks!

abolfazl-talaei:
@EtienneLem sorry for my late response. Yes the changes make sense and I think that's correct. Thank you.

PR #721, opened 2022-11-07
Remove restrictive custom emoji styling

### LINKED ISSUE #720: Seemingly unnecessary image styling in custom emojis inside picker
Due to the rigid styling decisions made inside the `Emoji.tsx` file here, custom emojis render with `position: relative` and `top: '.1em'`, making them not render in line with native emojis.

```
<span class="emoji-mart-emoji" data-emoji-set={props.set}>
  {imageSrc ? (
    <img
      style={{
        height: props.size || '1em',
        width: 'auto',
        display: 'inline-block',
        position: 'relative',
        top: '.1em',
      }}
      alt={emojiSkin.native || emojiSkin.shortcodes}
      src={imageSrc}
    />
  ) ...
```
Here's some sample output:

image

Given the css styling happens in the shadow-root, I'm unable to change the styling here, otherwise I happily would. I don't see a good reason to have these styles forced upon the end user, and am not sure if there's a reason they are there in the first place. Am I okay to submit a PR to remove them? At the very least could we add a custom styles prop to be passed in that sets the styles here?

thatvegandev:
Fixes #720

### DIFF
--- a/packages/emoji-mart/src/components/Emoji/Emoji.tsx
+++ b/packages/emoji-mart/src/components/Emoji/Emoji.tsx
@@ -42,8 +42,6 @@ export default function Emoji(props) {
             height: props.size || '1em',
             width: 'auto',
             display: 'inline-block',
-            position: 'relative',
-            top: '.1em',
           }}
           alt={emojiSkin.native || emojiSkin.shortcodes}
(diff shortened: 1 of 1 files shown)


EtienneLem:
Thanks!

PR #739, opened 2022-11-21
Support custom font size for Emoji Preview text

cmunozgar:
Set a custom font size for the Emoji Preview placeholder text. By default, `21px`.

CleanShot 2022-11-21 at 08 33 30@2x


packages/emoji-mart/src/components/Picker/Picker.tsx
```diff
@@ -702,7 +702,10 @@ export default class Picker extends Component {
                 </div>
               </div>
             ) : (
-              <div class="color-c" style={{ fontSize: 21 }}>
+              <div
+                class="color-c"
+                style={{ fontSize: this.props.emojiPreviewSize }}
```
guldenmw, line 707: I would recommend renaming this to something more clear, like `previewTextFontSize`.

packages/emoji-mart/src/components/Picker/PickerProps.ts
```diff
@@ -14,6 +14,9 @@ export default {
   emojiSize: {
     value: 24,
   },
+  emojiPreviewSize: {
+    value: 14,
```
guldenmw, line 18: Maybe keep the default 21px font size.

cmunozgar:
> Hey @cmunozgar 🙂
> 
> I left a couple of comments that I think would make this more clear.

Thanks for the suggestions @guldenmw

guldenmw (approved):

EtienneLem:
Thanks for the PR. I feel like it would make more sense as a CSS variable, so that it can be configured via CSS (i.e. https://missiveapp.com/open/emoji-mart/example-custom-styles.html).

Eventually, we will add [CSS Shadow Parts](https://www.w3.org/TR/css-shadow-parts-1) (see example https://css-tricks.com/styling-in-the-shadow-dom-with-css-shadow-parts) so that every part is easier to customize.

cmunozgar:
> Thanks for the PR. I feel like it would make more sense as a CSS variable, so that it can be configured via CSS (i.e. https://missiveapp.com/open/emoji-mart/example-custom-styles.html).
> 
> Eventually, we will add [CSS Shadow Parts](https://www.w3.org/TR/css-shadow-parts-1) (see example https://css-tricks.com/styling-in-the-shadow-dom-with-css-shadow-parts) so that every part is easier to customize.

Thank you, @EtienneLem, for your response!

I have simplified the PR to use the CSS variable instead, following the same naming convention. 

Let me know if this one makes more sense.

EtienneLem (approved):

EtienneLem:
Thank you @cmunozgar. I have updated it a bit:

- The `em-*` variables are meant to be internal; I added a few ones that will be more friendly to use.
- I added variables to control the font-size of the preview when hovering as well

It will be configurable like this:

```css
em-emoji-picker {
  --preview-placeholder-size: 14px;
}
```

cmunozgar:
Thank you @EtienneLem!

packages/emoji-mart/src/components/Picker/PickerStyles.scss
```diff
@@ -46,6 +46,10 @@
 
   --sidebar-width: 16px;
 
+  --preview-placeholder-size: 21px;
```
cmunozgar, line 49: What I realize here @EtienneLem is that we should probably use an internal variable name and check if a global variable is present or if the value should default to 21px. Otherwise, there's no way to customize the size.

cmunozgar: I've created this small PR to address the override of the CSS variable: https://github.com/missive/emoji-mart/pull/779

EtienneLem: Should be working in **v5.5.1**

PR #747, opened 2022-11-27
Update with fully Right to Left support for rtl languages

airani:
Add fully supports for Right to Left style and rtl languages

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,10 +7,19 @@
 </div>
 
 ## 📖 Table of Contents
+- [📖 Table of Contents](#-table-of-contents)
 - [💾 Data](#-data)
+  - [Bundled directly into your codebase](#bundled-directly-into-your-codebase)
+  - [Fetched remotely](#fetched-remotely)
 - [🏪 Picker](#-picker)
+  - [React](#react)
+  - [Browser](#browser)
+  - [Options / Props](#options--props)
+  - [Custom emojis](#custom-emojis)
+  - [Custom category icons](#custom-category-icons)
 - [🙃 Emoji component](#-emoji-component)
-- [🕵️‍♀️ Headless search](#%EF%B8%8F%EF%B8%8F-headless-search)
+  - [Attributes / Props](#attributes--props)
+- [🕵️‍♀️ Headless search](#️️-headless-search)
 - [🔬 Get emoji data from native](#-get-emoji-data-from-native)
 - [🗺 Internationalization](#-internationalization)
 - [📚 Examples](#-examples)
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -7,10 +7,19 @@
...
 - [🏪 Picker](#-picker)
+  - [React](#react)
+  - [Browser](#browser)
+  - [Options / Props](#options--props)
+  - [Custom emojis](#custom-emojis)
+  - [Custom category icons](#custom-category-icons)
 - [🙃 Emoji component](#-emoji-component)
-- [🕵️‍♀️ Headless search](#%EF%B8%8F%EF%B8%8F-headless-search)
+  - [Attributes / Props](#attributes--props)
+- [🕵️‍♀️ Headless search](#️️-headless-search)
```
EtienneLem, line 22: I would rather if we kept these kind of unrelated changes outside of this PR

README.md
```diff
@@ -110,6 +119,7 @@ function App() {
 | **perLine** | `9` | | The number of emojis to show per line |
 | **previewEmoji** | `point_up` | | The id of the emoji to use for the preview when not hovering any emoji. `point_up` when preview position is bottom and `point_down` when preview position is top. |
 | **previewPosition** | `bottom` | `top`, `bottom`, `none` | The position of the preview |
+| **rtl** | `false` | | Whether the Right to Left style or not |
```
EtienneLem, line 122: I’m not sure if we really need a prop for this 🤔. This should be controlled by the `i18n` data. Chances are there is not much use case of manually providing `rtl` without also providing a language that is rtl.

EtienneLem (changes requested):

packages/emoji-mart/src/components/Navigation/Navigation.tsx
```diff
@@ -54,7 +54,12 @@ export default class Navigation extends PureComponent {
     let selectedCategoryIndex = null
 
     return (
-      <nav id="nav" class="padding" data-position={this.props.position}>
+      <nav
+        id="nav"
+        class="padding"
+        data-position={this.props.position}
+        dir={this.props.dir}
```
EtienneLem, line 61: Should be `this.dir`

packages/emoji-mart/src/components/Navigation/Navigation.tsx
```diff
@@ -87,7 +92,10 @@ export default class Navigation extends PureComponent {
             style={{
               width: `${100 / this.categories.length}%`,
               opacity: selectedCategoryIndex == null ? 0 : 1,
-              transform: `translateX(${selectedCategoryIndex * 100}%)`,
+              transform:
+                this.props.dir === 'rtl'
```
EtienneLem, line 96: Should be `this.dir`

EtienneLem (changes requested):

packages/emoji-mart/src/components/Navigation/Navigation.tsx
airani: Are you sure? in this component we don't have `this.dir` and I sending `dir` as a props from Picker to Navigation

EtienneLem: Oh, my bad! I thought this was in Picker

airani:
@EtienneLem This PR is ok to merge?

EtienneLem (approved):

EtienneLem:
Thanks @airani!

PR #758, opened 2022-12-18
Relay event arg to `onClickOutside` callback

TwilightCoder:
This allows the callback to filter target elements.

### Sample issue scenario
Given:
- A custom trigger callback that opens the picker popup.
- And then the `onClickOutside` callback set to close the popup on outside clicks.

Issue: Without excluding the trigger element target in the `onClickOutside` callback, the popup is opened by the first callback and then closed instantly by the second callback, as it is registered as an outside click.


EtienneLem (approved):

EtienneLem:
Thanks! 🤘🏼

PR #767, opened 2023-01-04
fix: Picker width is changing when hover

### LINKED ISSUE #762: Bug: Broken truncation of Preview text
After release of v5.4.0 there is a problem with long emoji name in Preview. The text is not truncated but instead just changing width of the picker and causing jumpiness. 

Steps to reproduce:
1. Hover over on random emojis and see how Picker width is changing depending on the text in Preview.

Also, see gif.

ezgif com-gif-maker

rottenpen:
fix #762

### DIFF
--- a/packages/emoji-mart/src/components/Picker/Picker.tsx
+++ b/packages/emoji-mart/src/components/Picker/Picker.tsx
@@ -693,6 +693,11 @@ export default class Picker extends Component {
         id="preview"
         class="flex flex-middle"
         dir={this.dir}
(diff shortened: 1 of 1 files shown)


rottenpen:
@EtienneLem PTAL

EtienneLem:
Thanks for the reminder @rottenpen.

Your PR was breaking the layout:

Screenshot 2023-01-12 at 9 23 59 AM

I did find the right fix, it is handled in #772

PR #779, opened 2023-01-18
Support customization for Preview Placeholder Size

cmunozgar:
This PR enables the overwrite of the CSS variable `preview-placeholder-size`


EtienneLem:
Oh, yes these were meant to be set on `:host` to be configurable. Handled in #780.

PR #782, opened 2023-01-21
Add gujarati language set

Sidd-underscore:


Sidd-underscore:
ok all ready! sorry for confusing stuff above...

Issue #805, opened 2023-02-25
Problem in load custom svg emojies


mj-mehdizadeh:
I have some svg files that I want to load them as custom Emojies, but the `img` of svg doesn't show anything.

      <img id="svg-boombox" src="./assets/boombox.svg" /> // this images is exist and loaded
      <img id="img-shipit" src="./assets/shipit.png" /> // this is exist and loaded
```
       custom: [
          {
            emojis: [
              {
                id: 'octocat',
                name: 'Octocat',
                keywords: ['github'],
                skins: [{ src: document.querySelector('#svg-boombox').src }], // for this svg file, it doesn't show anything
              },
              {
                id: 'shipit',
                name: 'Squirrel',
                keywords: ['github'],
                skins: [{ src: document.querySelector('#img-shipit').src }], // for this png file, it's show the image
              },
            ],
          },
        ],
```

![image](https://user-images.githubusercontent.com/11937510/221347721-6797e0fc-8e99-4a19-b24b-45f60f4e48df.png)

mtojek:
Hi, I'm afraid this is still a valid issue. See below:

<img width="388" alt="Image" src="https://github.com/user-attachments/assets/dcbec412-f1fc-4cc1-9aa1-7c898d48932d" />

PR #806, opened 2023-02-28
load SVG images  

mj-mehdizadeh:
is this a best solution for this issue? #805


mtojek:
Hey @EtienneLem! We also confirmed the issue with SVGs on our side, the same root cause. Would you mind merging this PR?

PR #809, opened 2023-03-01
make noCountryFlags respect `false` on windows.

### LINKED ISSUE #808: `noCountryFlags: false` does still hide country flag emojis on windows
`noCountryFlags: false` does still hide flag emojis on ms windows with the `native` emoji set.
I replace the native emoji font inside of my css, with google noto color emoji which does support country flags:
```css
@font-face {
  font-family: 'EmojiMart';
  src: local('Apple Color Emoji'), url("noto/emoji/NotoColorEmoji.ttf") format("truetype");
}
```

But emoji mart checks only for the native font and does not notice that the `'EmojiMart'` font was set:

https://github.com/missive/emoji-mart/blob/d29728f7b4e295e46f9b64aa80335aa4a3c15b8e/packages/emoji-mart/src/helpers/native-support.ts#L64

### A. `noCountryFlags` property should respect false
noCountryFlags should autodetect when it is set to `undefined` and respect my preference when I set it to a boolean `true` or `false`:

```js
noCountryFlags = undefined // auto detect support
noCountryFlags = false // show country flags
noCountryFlags = true  // hide country flags
```

### B. `isEmojiSupported()` check should include `'EmojiMart'` font

probably easy to do (just add it to [native-support.ts#L64](https://github.com/missive/emoji-mart/blob/d29728f7b4e295e46f9b64aa80335aa4a3c15b8e/packages/emoji-mart/src/helpers/native-support.ts#L64)), but there is the question whether this introduces a race condition where the css stylesheet is loaded after emoji mart did the check.

Simon-Laux:
also fix a typo in the documentation

fixes #808 (this implements solution A)

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ function App() {
 | **locale** | `en` | `en`, `ar`, `be`, `cs`, `de`, `es`, `fa`, `fi`, `fr`, `hi`, `it`, `ja`, `kr`, `nl`, `pl`, `pt`, `ru`, `sa`, `tr`, `uk`, `vi`, `zh` | The locale to use for the picker |
 | **maxFrequentRows** | `4` | | The maximum number of frequent rows to show. `0` will disable frequent category |
 | **navPosition** | `top` | `top`, `bottom`, `none` | The position of the navigation bar |
-| **noCountryFlags** | `false` | | Whether to show country flags or not. If not provided, tbhis is handled automatically (Windows doesn’t support country flags) |
+| **noCountryFlags** | `undefined` | | Whether to show country flags or not. If not provided, this is handled automatically (Windows doesn’t support country flags) |
 | **noResultsEmoji** | `cry` | | The id of the emoji to use for the no results emoji |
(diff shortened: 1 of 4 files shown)


theavish (approved):

burgil (approved):
Looks great please approve this

ThenTech:
I see in your changes that this fix is after `if (set == 'native')`, but in my testing, if I use a different set, e.g. `google` and even have an emoji font specified, it also doesn't show the flags. But if I look at the online demo and select `google` there, the flags do show up...

Edit: I dug a bit deeper in the code, and it looks like you must specify the default set to use in the `init` function as well (which needs to be called if you want to render emojis with `em-emoji`). If I only give the data to `init`, it will default to `set="native"` and thus never actually use another icon set, even though the Picker has the correct `set` prop.
Using a different font face only does something if you use the `native` set, but that one then doesn't contain the flags in the first place (on Windows), so you must use a different set either way. So on Windows, providing a different font will never show flags.

PR #826, opened 2023-05-08
customization of styles

mukhrr:
Currently, customizing styles is kind of limited. User can change only em-emoji-picker.
This task includes customizing styles of other parts, including "Search" "Preview bar", "Underscore line of category tag icons" using [::part() CSS selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::part).
Also, added example on Custom styles part


stvncode:
Any update on this ? I want to remove the preview part but i don't know how to do it...

PR #868, opened 2023-09-26
[fix] matchMedia memory leak

kyle-w-20230331:
matchMedia may cause memory leak,  should removeListener when the component Unmount.


EtienneLem (approved):

EtienneLem:
Nice find, thank you!

PR #895, opened 2024-02-11
Add bengali localization

shahmedme:
Added Bengali language


akabutnicer (commented):
This was greatly needed, and fairly well made. I agree with this.

akabutnicer (approved):
But I do not think other developers are active anymore.

SalekurPolas (approved):

PR #898, opened 2024-02-25
locale ko -> kr 

cyd5538:
if you read the official documentation and install it,
npm install --save emoji-mart @emoji-mart/data @emoji-mart/react
version "emoji-mart": "^5.5.2" 

The Korean file is kr.json and the main.js locale choices is kr.
Because it is confusing, it would be better to consider kr.


packages/emoji-mart-website/index.html
```diff
@@ -75,7 +75,7 @@
             <option value="hi">हिन्दी</option>
             <option value="it">Italiano</option>
             <option value="ja">日本語</option>
```
oinochoe, line 77: @cyd5538 In that sense, I think “ja” should be changed to “jp”?

oinochoe: So I think it makes sense to consider "ko" or all the rest of the languages ​​also change.

cyd5538: @oinochoe  That's a good comment!

PR #935, opened 2024-07-30
Update README.md

msomotese:
Added a paragraph


PixsaOJ:
Love how people are not aware how shameful this is.

PR #988, opened 2025-03-30
docs: Add explanation on how to overwrite Data Store

### LINKED ISSUE #987: Make data store overridable again (for saving frequently used emojis and skin) + maybe make it asycronous?
## Old functionality

In the old pre-rewrite/pre-5.0 way, you could override the data store, 

```js
/**
 * Sets the emoji-mart data storage.
 *
 * @private
 * @returns {Promise<void>}
 * @see https://github.com/missive/emoji-mart#storage
 */
export async function initEmojiMartStorage() {
    // get saved values
    emojiMartStorage = await AddonSettings.get("emojiMart");

    globalThis.emojiMart.setDataStore({
        getter: (key) => {
            return emojiMartStorage[key];
        },

        setter: (key, value) => {
            emojiMartStorage[key] = value;

            // and actually save the new value
            // and return promise so async things are noticed
            return AddonSettings.set("emojiMart", emojiMartStorage);
        }
    });
}
```

taken [from here](https://github.com/rugk/awesome-emoji-picker/blob/744305167b94d3d5f77cb550bc21b7137fe08a38/src/popup/module/EmojiSettings.js#L34-L57).

See https://github.com/missive/emoji-mart/tree/v2.10.0?tab=readme-ov-file#storage for the old documentation. It said:
> By default EmojiMart will store user chosen skin and frequently used emojis in localStorage.

## New v5 implementation
Now looking for the same functionality in v5 I am a little lost [but searching through](https://github.com/search?q=repo%3Amissive%2Femoji-mart+store&type=code) [the code](https://github.com/search?q=repo%3Amissive%2Femoji-mart%20data%20store&type=code) I can see at least the skin is still read from it:
https://github.com/missive/emoji-mart/blob/16978d04a766eec6455e2e8bb21cd8dc0b3c7436/packages/emoji-mart/src/components/Picker/Picker.tsx#L33

[And frequently used, too, so hmm](https://github.com/missive/emoji-mart/blob/16978d04a766eec6455e2e8bb21cd8dc0b3c7436/packages/emoji-mart/src/helpers/frequently-used.ts). But I am unsure and there is no doc on how to overwrite the storage method. I doubt this is really possible at all, especially in the compiled versions.

rugk:
Fixes https://github.com/missive/emoji-mart/issues/987

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ Store.get = (key) => {
 /**
  * Stores a value under a specific key.
  * @param {string} key - The key to store the value under.
- * @param {string} value - The value to store.
+ * @param {any} value - The value to store.
  */
 Store.set = (key, value) => {
     console.info("Saving", value, `in '${key}'`);
--- a/README.md
+++ b/README.md
@@ -271,6 +271,34 @@ new Picker({ i18n })
 
 Given the small file size, English is built-in and doesn’t need to be provided.
 
+## 💾 Data Store
+By default EmojiMart will store user chosen skin and frequently used emojis in localStorage. That can however be overwritten if you can overwrite imports, should you want to store these in your own storage:
+
+```js
+import { Store } from 'emoji-mart'
+
+/**
+ * Retrieves a stored value by key.
+ * @param {string} key - The key of the stored value.
+ * @returns {any} The retrieved value.
+ */
+Store.get = (key) => {
+    console.info(`Value was requested for '${key}'`);
+    // Get from your own storage (sync)
+    return // ...
+};
+
+/**
+ * Stores a value under a specific key.
+ * @param {string} key - The key to store the value under.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -271,6 +271,34 @@ new Picker({ i18n })
 
 Given the small file size, English is built-in and doesn’t need to be provided.
 
+## 💾 Data Store
+By default EmojiMart will store user chosen skin and frequently used emojis in localStorage. That can however be overwritten if you can overwrite imports, should you want to store these in your own storage:
```
rugk:
```suggestion
By default EmojiMart will store user chosen skin and frequently used emojis in `localStorage`. That can however be overwritten if you can overwrite imports, should you want to store these in your own storage:
```

Issue #576, opened 2022-04-27
Update types over at @types/emoji-mart for v5


Scalahansolo:
I might be able to find some time to work through this. But... I figured a maintainer would be able to blow this considerably faster.

More than anything, I wanted to a tracking ticket open in this repo for tracking purposes.

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/emoji-mart

EtienneLem:
Indeed! It would most likely be more convenient if we were to include types right here, right? It would be easier to maintain at the source I assume.

Scalahansolo:
It would definitely be nice to ship releases of EmojiMart 5+ with a `.d.ts` file along with the release as to not have to install the disconnected dep from @types/emoji-mart

Scalahansolo:
Another option here would be to just convert the codebase to TypeScript.... If that's something y'all would be open too, I could make do that work. But would want to make sure the Missive crew would be okay with that before I put in the time.

EtienneLem:
Our Parcel scripts have the `--no-autoinstall` flag, you most likely need to add the `@parcel/transformer-typescript-types` to `packages/emoji-mart/package.json`

Scalahansolo:
![ScreenShot 04-27 at 11 52](https://user-images.githubusercontent.com/4317253/165589061-4a025c24-a280-46b8-ab73-c127db3d7cbc.png)

I gave that a shot and still running into the same issue.

Also tried a diff plugin based on their docs in a .parcelrc, no luck there either 

![ScreenShot 04-27 at 11 53](https://user-images.githubusercontent.com/4317253/165589186-14a4952d-3947-4b86-b567-754d228d2973.png)

EtienneLem:
Yeah I see, I’m getting the same error on `yarn build`. There must be something specific about `types` in Parcel codebase. From what I can see the `@parcel/transformer-typescript-types` does get properly installed in `root/node_modules` (as is expected)

Scalahansolo:
On my local branch, I actually made some good progress on converting the whole lib to typescript. If we could get past this build issue / get y'alls blessing on move the project to typescript, I would be happy to finish that work out.

zhouzi:
@Scalahansolo could you open a pull request with your progress? I am looking forward to the types being added and having your code would make it easier to help with the build issue 🙏

Scalahansolo:
[...]

PR #995, opened 2025-05-12
Add simple TS typing for Emoji.d.ts

rugk:
Ref https://github.com/missive/emoji-mart/issues/576


rugk:
Ah sotrry, just now found the correct typing already exists, but in emoji-mart-data: Superseded by https://github.com/missive/emoji-mart/pull/996

PR #996, opened 2025-05-12
Fix index.d.ts by adding missing typings

rugk:
* `shortcodes` was missing from the skins
* I also found `search` 

Both found by analyzing the response from `SearchIndex.search`. Maybe this method however just adds/modifies the types?

Ref https://github.com/missive/emoji-mart/issues/576

### DIFF
--- a/packages/emoji-mart-data/index.d.ts
+++ b/packages/emoji-mart-data/index.d.ts
@@ -17,6 +17,7 @@ export interface Emoji {
   skins: Skin[]
   version: number
   emoticons?: string[]
+  search?: string
 }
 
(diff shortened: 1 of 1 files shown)


rugk:
Apparently, the shortcode is just the ID of the emoji set with added colons (`:`) so (and optionally appended by the skin modifier):
https://github.com/missive/emoji-mart/blob/16978d04a766eec6455e2e8bb21cd8dc0b3c7436/packages/emoji-mart/src/config.ts#L245

SvenBudak:
Any news on this? i really hope for a merge :)

